PR #5, opened 2021-12-01
MYC-106

BtencateSphereon:


mehmetbicak (approved):
Based on the test I am able to login with SIOP, based on DIDs that are known to the Veramo Agent.

packages/did-auth-siop-op-authenticator/README.md
```diff
@@ -0,0 +1,27 @@
...
+  <br>
+</h1>
+
+---
+
+**Warning: This package still is in every early development. Breaking changes without notice will happen at this point!**
+
+---
+
+A Veramo plugin to authenticate using the Self Issued OpenID Provider v2 (SIOPv2)
```
nklomp, line 15: Could we add a bit more documentation and link to our SIOP README, which is quite extensive?

Ideally this README refers to the SIOP README for most and then lists the differences and/or things to consider in the context of Veramo. Also have some example code and calls would be helpful

packages/did-auth-siop-op-authenticator/__tests__/restAgent.test.ts
```diff
@@ -0,0 +1,75 @@
...
+let serverAgent: IAgent
+let restServer: Server
+
+const getAgent = (options?: IAgentOptions) =>
+  createAgent<IDidAuthSiopOpAuthenticator>({
+    ...options,
+    plugins: [
+      new DidAuthSiopOpAuthenticator({
+        did: 'did:ethr:0xcBe71d18b5F1259faA9fEE8f9a5FAbe2372BE8c9',
+        kid: 'did:ethr:0xcBe71d18b5F1259faA9fEE8f9a5FAbe2372BE8c9',
```
nklomp, line 26: Thats no KID it seems, that is the DID

packages/did-auth-siop-op-authenticator/__tests__/shared/didAuthSiopOpAuthenticatorAgentLogic.ts
```diff
@@ -0,0 +1,243 @@
...
+      mockedSelectVerifiableCredentialsForSubmissionMethod.mockReturnValue(Promise.resolve({ errors: [], matches: ['match'] }))
+
+      const mockedSubmissionFromMethod = jest.fn()
+      PresentationExchange.prototype.submissionFrom = mockedSubmissionFromMethod
+      mockedSubmissionFromMethod.mockReturnValue(Promise.resolve({}))
+    })
+
+    afterAll(testContext.tearDown)
+
+    it('should authentication with DID SIOP without custom approval', async () => {
```
nklomp, line 143: -> authenticate

packages/did-auth-siop-op-authenticator/__tests__/shared/didAuthSiopOpAuthenticatorAgentLogic.ts
```diff
@@ -0,0 +1,243 @@
...
+      const result = await agent.authenticateWithDidSiop({
+        stateId,
+        redirectUrl,
+        didMethod,
+      })
+
+      expect(result.status).toEqual(200)
+    })
+
+    it('should authentication with DID SIOP with custom approval', async () => {
```
nklomp, line 153: -> authenticate

That is true for all it/test strings

packages/did-auth-siop-op-authenticator/package.json
```diff
@@ -0,0 +1,48 @@
...
+  },
+  "repository": "git@github.com:Sphereon-Opensource/SSI-SDK.git",
+  "author": "Sphereon <dev@sphereon.com>",
+  "license": "Apache-2.0",
+  "keywords": [
+    "Sphereon",
+    "SSI",
+    "Veramo",
+    "DID",
+    "SIOP",
```
nklomp, line 45: Add: SIOPv2, Presentation Exchange, OIDC4VP, OpenID Connect

packages/did-auth-siop-op-authenticator/src/agent/DidAuthSiopOpAuthenticator.ts
```diff
@@ -0,0 +1,204 @@
...
+  private readonly op: OP
+  private readonly did: string
+  private readonly kid: string
+  private readonly privateKey: string
+  private readonly expiresIn: number | undefined
+  private readonly didMethod: string | undefined
+  private readonly customApprovals: Record<string, (verifiedAuthenticationRequest: VerifiedAuthenticationRequestWithJWT) => Promise<void>>
+
+  constructor(options: IDidAuthSiopOpAuthenticatorArgs) {
+    this.did = options.did
```
nklomp, line 53: Hmzz, I would expect Veramo identifiers to be used, instead of DID strings, same for KID and privateKey

packages/did-auth-siop-op-authenticator/src/agent/DidAuthSiopOpAuthenticator.ts
```diff
@@ -0,0 +1,204 @@
...
+        return {
+          location: presentationDef.location,
+          format: VerifiablePresentationTypeFormat.LDP_VP,
+          presentation: verifiablePresentation,
+        }
+      })
+    )
+  }
+
+  private getPresentationExchange(verifiableCredentials: VerifiableCredential[]): PresentationExchange {
```
nklomp, line 155: This would need to be integrated with Veramo Verifiable Credentials and Storage, but that one can wait for now I guess

packages/did-auth-siop-op-authenticator/src/agent/DidAuthSiopOpAuthenticator.ts
```diff
@@ -0,0 +1,204 @@
...
+    })
+  }
+
+  /** {@inheritDoc IDidAuthSiopOpAuthenticator.verifyDidSiopAuthenticationRequestURI} */
+  private async verifyDidSiopAuthenticationRequestURI(
+    args: IVerifyDidSiopAuthenticationRequestUriArgs,
+    context: IRequiredContext
+  ): Promise<VerifiedAuthenticationRequestWithJWT> {
+    const didMethodsSupported = args.requestURI.registration?.did_methods_supported as string[]
+    let didMethods: string[] = []
```
nklomp, line 168: Hmzz, what we are doing here is the RP is stating its supported DIDs. We are now adding that to our Veramo/RP list and resolver. What we should be doing is verifying it against locally (RP/Veramo) supported DIDs first and only choose those that we actually support. As long as we use our Universal DID client in the Agent we should be good for now, but it is wise to add a fixme and/or ticket here

nklomp (changes requested):
I am missing the integration with Veramo itself. This now acts more like a wrapper to SIOP.

The idea is that you have Identifiers (DIDs) and keys (kids) of these DIDs in Veramo. 
Same for VCs, although we can skip that integration for now. I would expect the Identifier to be part of the method arguments for instance. It would then look them up from Veramo's Key Management System and DID manager

packages/did-auth-siop-op-authenticator/src/agent/DidAuthSiopOpAuthenticator.ts
nklomp: Ideally what should happen here is that we expose which identifiers are potential matches to the did_methods_supported.
The RP lists which DID methods are supported. The DIDManager knows which DIDs it has. So it could filter the list to only the DIDs that have a DID method from the did_methods_supported. If dids_methods_supported is not provided/empty it means all DIDs could be used

packages/did-auth-siop-op-authenticator/README.md
BtencateSphereon: added a better readme

packages/did-auth-siop-op-authenticator/__tests__/restAgent.test.ts
BtencateSphereon: should be a kid now

packages/did-auth-siop-op-authenticator/__tests__/shared/didAuthSiopOpAuthenticatorAgentLogic.ts
BtencateSphereon: fixed

packages/did-auth-siop-op-authenticator/package.json
BtencateSphereon: added, also kept SIOP, not sure if you wanted SIOPv2 instead of SIOP, but now they are both in there

packages/did-auth-siop-op-authenticator/src/agent/DidAuthSiopOpAuthenticator.ts
BtencateSphereon: So I tried using the Veramo credentials from the datastore, but they are incompatible with PE-JS. So we need to do some mapping or start supporting them. 

For now I added a TODO with a link to a ticket

packages/did-auth-siop-op-authenticator/src/agent/DidAuthSiopOpAuthenticator.ts
BtencateSphereon: Added a todo with a ticket link

package.json
```diff
@@ -8,7 +8,7 @@
   "license": "Apache-2.0",
   "scripts": {
     "build": "lerna run build && yarn build:schema-api",
-    "build:schema-api": "lerna run extract-api && lerna run generate-plugin-schema",
+    "build:schema-api": "lerna run --ignore @sphereon/ssi-sdk-did-auth-siop-authenticator extract-api && lerna run --ignore @sphereon/ssi-sdk-did-auth-siop-authenticator generate-plugin-schema",
```
nklomp, line 11: Have you tested what happens if we remove the build:schema-api step from the respective packages/<name>/package.json? Or to have a command that basically does nothing there? I think that to be a better place to define that no schema generation should be done, than in the parent workspace package.json

packages/did-auth-siop-op-authenticator/README.md
```diff
@@ -0,0 +1,214 @@
...
+<h1 align="center">
+  <br>
+  <a href="https://www.sphereon.com"><img src="https://sphereon.com/content/themes/sphereon/assets/img/logo.svg" alt="Sphereon" width="400"></a>
+  <br>DID Auth SIOP OP Authenticator (Typescript) 
+  <br>
+</h1>
+
+---
+
+**Warning: This package still is in every early development. Breaking changes without notice will happen at this point!**
```
nklomp, line 11: every -> very 😂 Might be true though ;)

packages/did-auth-siop-op-authenticator/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+---
+
+A Veramo operating party authentication plugin using the [Self Issued OpenID Provider v2 (SIOP)](https://github.com/Sphereon-Opensource/did-auth-siop) authentication library for having clients / people conforming to
```
nklomp, line 15: operating party? I think you are confusing a Self issued OpenID Provider (SIOP) with a Relying Party. Currently the plugin is mainly involved on the SIOP side of things.
Given the complete sentence I think it makes most sense to just have 'A Veramo authentication ....'

packages/did-auth-siop-op-authenticator/README.md
```diff
@@ -0,0 +1,214 @@
...
+* getDidSiopAuthenticationRequestFromRP
+* getDidSiopAuthenticationRequestDetails
+* verifyDidSiopAuthenticationRequestURI
+* sendDidSiopAuthenticationResponse
+
+## Usage
+
+### Adding the plugin to an agent:
+
+```js
```
nklomp, line 52: typescript is supported as language

packages/did-auth-siop-op-authenticator/README.md
```diff
@@ -0,0 +1,214 @@
...
+import { IDidAuthSiopOpAuthenticator } from '@sphereon/ssi-sdk-did-auth-siop-authenticator'
+
+const agent = createAgent<IDidAuthSiopOpAuthenticator & IResolver>({
+  plugins: [
+    new DidAuthSiopOpAuthenticator(),
+    new DIDResolverPlugin({
+      resolver: new Resolver({
+        ...ethrDidResolver({ infuraProjectId: INFURA_PROJECT_ID }),
+        ...webDidResolver(),
```
nklomp, line 61: I would like to include our UniResilver in the examples. Also a didkey makes more sense then the ethrDidResolver propbably, unless we can ommit the INFURA stuff. We shouldn't be needing that anyway, since basic DIDs using the ethr driver do not need actual on chain state

packages/did-auth-siop-op-authenticator/agent.yml
```diff
@@ -0,0 +1,107 @@
+version: 3.0
+
+constants:
+  baseUrl: http://localhost:3335
+  port: 3335
+  # please use your own X25519 key, this is only an example
+  secretKey: 29739248cad1bd1a0fc4d9b75cd4d2990de535baf5caadfdf8d8f86664aa830c
+  methods:
+    - resolveDid
```
nklomp, line 9: Are these all methods? Why do we expose a resolveDid, whilst we also rely on the DID plugin already, which exposes something similar?

packages/did-auth-siop-op-authenticator/package.json
```diff
@@ -0,0 +1,53 @@
...
+  },
+  "dependencies": {
+    "@sphereon/did-auth-siop": "^0.2.2",
+    "@sphereon/pe-js": "^0.5.0",
+    "@veramo/core": "^3.1.0",
+    "@veramo/utils": "^3.1.1-next.17",
+    "cross-fetch": "^3.1.4"
+  },
+  "devDependencies": {
+    "@veramo/cli": "^3.1.0",
```
nklomp, line 24: Although probably not problematic, makes sense to use the same version as the other @veramo packages, given the whole SSI-SDK will need to use the same versions of upstream plugins

packages/did-auth-siop-op-authenticator/package.json
```diff
@@ -0,0 +1,53 @@
...
+    }
+  },
+  "scripts": {
+    "build": "tsc --build",
+    "generate-plugin-schema": "yarn veramo dev generate-plugin-schema"
+  },
+  "dependencies": {
+    "@sphereon/did-auth-siop": "^0.2.2",
+    "@sphereon/pe-js": "^0.5.0",
+    "@veramo/core": "^3.1.0",
```
nklomp, line 19: Although probably not problematic, makes sense to use the same version as the other @veramo packages, given the whole SSI-SDK will need to use the same versions of upstream plugins

packages/did-auth-siop-op-authenticator/package.json
```diff
@@ -0,0 +1,53 @@
...
+      "IDidAuthSiopOpAuthenticator": "./src/types/IDidAuthSiopOpAuthenticator.ts"
+    }
+  },
+  "scripts": {
+    "build": "tsc --build",
+    "generate-plugin-schema": "yarn veramo dev generate-plugin-schema"
+  },
+  "dependencies": {
+    "@sphereon/did-auth-siop": "^0.2.2",
+    "@sphereon/pe-js": "^0.5.0",
```
nklomp, line 18: 0.5.1 of PE-JS is released with some critical fixes we need for this integration to actually start working in creating VPs in production scenarios later on

packages/did-auth-siop-op-authenticator/src/agent/DidAuthSiopOpAuthenticator.ts
```diff
@@ -0,0 +1,166 @@
...
+    authenticateWithDidSiop: this.authenticateWithDidSiop.bind(this),
+    getDidSiopAuthenticationRequestFromRP: this.getDidSiopAuthenticationRequestFromRP.bind(this),
+    getDidSiopAuthenticationRequestDetails: this.getDidSiopAuthenticationRequestDetails.bind(this),
+    verifyDidSiopAuthenticationRequestURI: this.verifyDidSiopAuthenticationRequestURI.bind(this),
+    sendDidSiopAuthenticationResponse: this.sendDidSiopAuthenticationResponse.bind(this),
+    registerCustomApprovalForDidSiop: this.registerCustomApprovalForDidSiop.bind(this),
+    removeCustomApprovalForDidSiop: this.removeCustomApprovalForDidSiop.bind(this)
+  }
+
+  private readonly sessions: Record<string, OperatingPartySession>
```
nklomp, line 43: Just rename to OPSession, as it stands for OpenId Provider not Operating Party. OP is a term well known in the OpenID world

packages/did-auth-siop-op-authenticator/src/agent/DidAuthSiopOpAuthenticator.ts
```diff
@@ -0,0 +1,166 @@
...
+    this.customApprovals = customApprovals || {}
+  }
+
+  /** {@inheritDoc IDidAuthSiopOpAuthenticator.getSession} */
+  private async getDidSiopSession(
+      args: IGetDidSiopSessionArgs,
+      context: IRequiredContext
+  ): Promise<OperatingPartySession> {
+    if (this.sessions[args.sessionId] === undefined) {
+      return Promise.reject(new Error(`No session found for id: ${args.sessionId}`))
```
nklomp, line 57: I think we might at a TODO to cleanup the sessions in the and. Probably using a configurable cache makes sense here

packages/did-auth-siop-op-authenticator/src/agent/DidAuthSiopOpAuthenticator.ts
```diff
@@ -0,0 +1,166 @@
...
+    }
+
+    return this.sessions[args.sessionId]
+  }
+
+  /** {@inheritDoc IDidAuthSiopOpAuthenticator.addSessionForDidSiop} */
+  private async addSessionForDidSiop(
```
nklomp, line 64: I think we should allow for no sessionId to be passed in as well in which case we just generate a random sessionId. Probably makes sense to rename the method to 'new' instead of 'add' then as well

packages/did-auth-siop-op-authenticator/src/agent/DidAuthSiopOpAuthenticator.ts
```diff
@@ -0,0 +1,166 @@
...
+    const sessionId = args.sessionId
+    const session = new OperatingPartySession({sessionId, identifier: args.identifier, expiresIn: args.expiresIn, context})
+    await session.init()
+    this.sessions[sessionId] = session
+
+    return session
+  }
+
+  /** {@inheritDoc IDidAuthSiopOpAuthenticator.removeSessionForDidSiop} */
+  private async removeSessionForDidSiop(
```
nklomp, line 81: I know we are using DidSiop in the SIOP library a lot as well, but propose to just rename all of these methods to omit the Did part, as it adds nothing. So this one would become removeSessionForSiop

packages/did-auth-siop-op-authenticator/src/session/OperatingPartySession.ts
```diff
@@ -0,0 +1,231 @@
...
+  IOpsSendDidSiopAuthenticationResponseArgs,
+  IOpsVerifyDidSiopAuthenticationRequestUriArgs,
+  IAuthRequestDetails,
+  IMatchedPresentationDefinition,
+  IRequiredContext,
+} from '../types/IDidAuthSiopOpAuthenticator'
+
+const fetch = require('cross-fetch')
+
+export class OperatingPartySession {
```
nklomp, line 30: OPSession

packages/did-auth-siop-op-authenticator/src/session/OperatingPartySession.ts
```diff
@@ -0,0 +1,231 @@
...
+  public readonly id: string
+  public readonly identifier: IIdentifier
+  public readonly didMethod: string
+  public readonly section: DIDDocumentSection
```
nklomp, line 34: didDocSection or verificationMethodSection?

packages/did-auth-siop-op-authenticator/src/utils/index.ts
```diff
@@ -0,0 +1,51 @@
+import { IParsedDID } from '../types/IDidAuthSiopOpAuthenticator';
```
nklomp, line 1: I think it might make sense to move this to the core plugin

nklomp (changes requested):
Although there a quite a few comments, they are all minor I believe, so good work!

package.json
BtencateSphereon: yep that also works, changed

packages/did-auth-siop-op-authenticator/README.md
BtencateSphereon: fixed

packages/did-auth-siop-op-authenticator/README.md
BtencateSphereon: ye makes sense, changed

packages/did-auth-siop-op-authenticator/README.md
BtencateSphereon: fixed

packages/did-auth-siop-op-authenticator/README.md
BtencateSphereon: changed

packages/did-auth-siop-op-authenticator/agent.yml
BtencateSphereon: wasnt needed, removed

packages/did-auth-siop-op-authenticator/package.json
BtencateSphereon: i do not see any differences, we use 3.1.0 everywhere

packages/did-auth-siop-op-authenticator/package.json
BtencateSphereon: i do not see any differences, we use the same version everywhere

packages/did-auth-siop-op-authenticator/package.json
BtencateSphereon: did-auth-siop is not yet compatible with pe-js 0.5.1

packages/did-auth-siop-op-authenticator/src/agent/DidAuthSiopOpAuthenticator.ts
BtencateSphereon: added a TODO with a link to a ticket

packages/did-auth-siop-op-authenticator/src/agent/DidAuthSiopOpAuthenticator.ts
BtencateSphereon: sessionId is now optional and we will generate an uuid for if one is not provided

packages/did-auth-siop-op-authenticator/src/agent/DidAuthSiopOpAuthenticator.ts
BtencateSphereon: changed the names to omit did

packages/did-auth-siop-op-authenticator/src/session/OperatingPartySession.ts
BtencateSphereon: renamed

packages/did-auth-siop-op-authenticator/src/session/OperatingPartySession.ts
BtencateSphereon: renamed to verificationMethodSection

packages/did-auth-siop-op-authenticator/src/utils/index.ts
BtencateSphereon: moved it to the core plugin

packages/did-auth-siop-op-authenticator/package.json
nklomp: A new version 0.2.3 is just released

packages/did-auth-siop-op-authenticator/package.json
nklomp: "@veramo/utils": "^3.1.1-next.17",

packages/did-auth-siop-op-authenticator/package.json
nklomp: "@veramo/utils": "^3.1.1-next.17",

packages/did-auth-siop-op-authenticator/src/agent/DidAuthSiopOpAuthenticator.ts
nklomp: Makes sense

nklomp (changes requested):
Basically only an update of deps left

packages/did-auth-siop-op-authenticator/package.json
BtencateSphereon: implemented

packages/did-auth-siop-op-authenticator/package.json
BtencateSphereon: there is no @veramo/utils 3.1.0, seems they are still working on their initial release

nklomp (changes requested):
LGTM, except there is a yarn.lock conflict, probably because something else got merged

nklomp (approved):

PR #6, opened 2021-12-07
MYC-140 fix prettier errors

hrehman-sphereon:
The develop already has a test failure. The test failuer is not because of the refactoring.


nklomp:
You might have a different setup, as develop is building in CI. Nonetheless good to update it anyway

nklomp (approved):

PR #7, opened 2021-12-07
MYC-140 Add to readme required environment variable information & upd…

hrehman-sphereon:
MYC-140 Add to readme required environment variable information & update .gitignore
Though actually the Agent should have been mocked properly.

### DIFF
--- a/packages/vc-api-verifier/README.md
+++ b/packages/vc-api-verifier/README.md
@@ -29,6 +29,22 @@ yarn build
 ```
 
 ### Test
+To run the tests a `VC_HTTP_API_AUTH_TOKEN` environment variable needs to be present with an authorization token.
+
+#### Environment Variable:
+
+- Navigate
+    - `ssi-sdk/package.json`
+    - `scripts.test`
+- Right click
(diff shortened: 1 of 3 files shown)


packages/vc-api-verifier/README.md
```diff
@@ -29,6 +29,22 @@ yarn build
 ```
 
 ### Test
+To run the tests a `VC_HTTP_API_AUTH_TOKEN` environment variable needs to be present with an authorization token.
+
+#### Environment Variable:
```
nklomp, line 34: These are realy specific notes that will certainly not apply to all environment. I think it would even confuse people. We shouldn't be explaining how to set environment variables, just that they are needed

nklomp (changes requested):
See remarks. I think the env remarks are way to specific, and will only confuse a lot of people using the software, as it seems to take Intelij as a precondition without mentioning that.

We shouldn't be telling people how to set env variables, just that they are needed

hrehman-sphereon:
As per comments it is not required.

PR #8, opened 2021-12-10
MYC-138 Create seed/wallet plugin

maikel-maas:
- Integrated this project within the monorepo;

### DIFF
--- a/packages/mnemonic-info-generator/src/agent/MnemonicInfoGenerator.ts
+++ b/packages/mnemonic-info-generator/src/agent/MnemonicInfoGenerator.ts
@@ -1,0 +1,106 @@
+import * as crypto from 'crypto';
+
+import { IAgentPlugin } from '@veramo/core';
+import { AbstractSecretBox } from '@veramo/key-manager';
+import * as bip39 from 'bip39';
+import { Connection } from 'typeorm';
+
+import {
+  DeleteResult,
+  IMnemonicGeneratorArgs,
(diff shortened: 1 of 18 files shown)


packages/mnemonic-info-generator/src/agent/MnemonicInfoGenerator.ts
```diff
@@ -37,7 +37,7 @@ export class MnemonicInfoGenerator implements IAgentPlugin {
 
   private async generateMnemonic(args: IMnemonicGeneratorArgs): Promise<IMnemonicInfoResult> {
     const mnemonic = bip39.generateMnemonic(args.bits);
-    if (args.shouldSave) {
+    if (args.shouldPersist) {
```
nklomp, line 42: I mentioned to remove the word should, because it suggests outcome optionality

nklomp (commented):
The important part of my comment about save -> persist was the removal of 'should', as 'should' suggests outcome optionality, which is really not there. Simply name it 'persist'.

If I say 'should save', then somebody still cannot be sure that I actually saved it

Rest looks okay

packages/mnemonic-info-generator/src/agent/MnemonicInfoGenerator.ts
maikel-maas: If it's ok for you I will change the name to "persist" only.

nklomp (approved):

PR #30, opened 2022-04-29
Feature/vdp 27 adding bbs to jsonld suites

maikel-maas:

PR #33, opened 2022-05-19
Feature/vdp 47

maikel-maas:


nklomp (commented):
I am confused. This seems to be the same or a lot of overlap with https://github.com/Sphereon-Opensource/ssi-sdk/pull/30 Can you clarify?

packages/vc-handler-ld-local/src/__tests__/shared/vcHandlerLocalAgentBbsLogic.ts
```diff
@@ -0,0 +1,206 @@
...
+        })
+      )
+    })
+
+    it('Should verify a BBS+ verifiable credential', async () => {
+      await expect(agent.verifyCredentialLDLocal({ credential: verifiableCredential })).resolves.toEqual(true)
+    })
+
+    it('Should create a BBS+ verifiable presentation', async () => {
+      const presentationPayload: PresentationPayload = {
```
maikel-maas, line 141: This one should be reviewed.

packages/vc-handler-ld-local/src/__tests__/shared/vcHandlerLocalAgentBbsLogic.ts
```diff
@@ -0,0 +1,206 @@
...
+                verificationMethod: expect.any(String),
+              },
+              type: ['VerifiableCredential', 'PermanentResidentCard'],
+            },
+          ],
+        })
+      )
+    })
+
+    it('Should verify a BBS+ verifiable presentation', async () => {
```
maikel-maas, line 202: This one should be reviewed.

maikel-maas (commented):
There are only two methods to review create VP and verify VP. Since VDP-27 isn't merged the code will look very similar.

packages/vc-handler-ld-local/src/ld-credential-module.ts
```diff
@@ -55,18 +59,29 @@ export class LdCredentialModule {
...
-      purpose,
-      suite: signingSuite,
-      documentLoader,
-      compactProof: false,
-    })
+
+    let verifiableCredential
+    //Needs to be signed using jsonld-signaures@5.0.1
+    if (key.type === 'Bls12381G2') {
+      verifiableCredential = await jsonld.sign(credential, {
```
maikel-maas, line 71: To be reviewed

packages/vc-handler-ld-local/src/ld-credential-module.ts
```diff
@@ -147,17 +182,30 @@ export class LdCredentialModule {
...
-      documentLoader: this.ldDocumentLoader.getLoader(context, fetchRemoteContexts),
-      challenge,
-      domain,
-      presentationPurpose,
-      compactProof: false,
-      checkStatus,
-    })
+    let result
+    if (presentation.proof.type === 'BbsBlsSignature2020') {
+      //Should never be null or undefined
```
maikel-maas, line 187: to review

packages/vc-handler-ld-local/src/ld-credential-module.ts
```diff
@@ -91,6 +106,13 @@ export class LdCredentialModule {
 
     suite.preSigningPresModification(presentation)
 
+    if (key.type === 'Bls12381G2') {
+      return await jsonld.sign(presentation, {
```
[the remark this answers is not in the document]
nklomp: vc-js signPresentation has some additional logic. Which means we are not symetric in our handling of BLS signatures. https://github.com/digitalcredentials/vc-js/blob/9e40e67a3060289c899890c2cfb76681cb893368/lib/vc.js#L354

packages/vc-handler-ld-local/src/ld-credential-module.ts
```diff
@@ -111,15 +133,28 @@ export class LdCredentialModule {
...
-    })
-
+    let result
+    if (credential.proof.type === 'BbsBlsSignature2020') {
+      //Should never be null or undefined
+      const suite = this.ldSuiteLoader
+        .getAllSignatureSuites()
+        .find((s) => s.getSupportedVeramoKeyType() === 'Bls12381G2')
+        ?.getSuiteForVerification() as BbsBlsSignature2020
+      result = await jsonld.verify(credential, {
```
nklomp, line 143: This misses stuff like revocationStatus and is not the same as vc.verifyCredential
https://github.com/digitalcredentials/vc-js/blob/9e40e67a3060289c899890c2cfb76681cb893368/lib/vc.js#L256

packages/vc-handler-ld-local/src/ld-credential-module.ts
nklomp: Not the same as vc.verify https://github.com/digitalcredentials/vc-js/blob/9e40e67a3060289c899890c2cfb76681cb893368/lib/vc.js#L397

nklomp (changes requested):
Okay, I didn't really notice it before. the signatures package only checks and generates signatures. The fact that returns VCs or has VCs as input, doesn't mean you can simply bypass VC-JS. VC-JS does additional checks, has some logic and for instance adds status support to credentials. The end result is that we now lack certain features for BLS VCs and that the error checking is not the same.

Is there a specific reason why we simply couldn't use VC-JS to begin with? It depends on the same package you are now using directly anyway

packages/vc-handler-ld-local/src/ld-credential-module.ts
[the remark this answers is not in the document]
maikel-maas: Cannot use it because vc-js uses different packages

require('@digitalcredentials/jsonld');
const jsigs = require('@digitalcredentials/jsonld-signatures');

packages/vc-handler-ld-local/src/ld-credential-module.ts
maikel-maas: Cannot use it because vc-js uses different packages

require('@digitalcredentials/jsonld');
const jsigs = require('@digitalcredentials/jsonld-signatures');

packages/vc-handler-ld-local/src/ld-credential-module.ts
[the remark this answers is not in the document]
nklomp: Yes? Those are forks compatible with RN

packages/vc-handler-ld-local/src/ld-credential-module.ts
nklomp: Yes these those are forks of the same packages that can run on RN as well.

Obviosuly we cannot end up with missing pieces and different logic

packages/vc-handler-ld-local/src/ld-credential-module.ts
maikel-maas: Cannot use it because vc-js uses different packages

require('@digitalcredentials/jsonld');
const jsigs = require('@digitalcredentials/jsonld-signatures');

packages/vc-handler-ld-local/src/ld-credential-module.ts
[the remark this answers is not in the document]
maikel-maas: Cannot use it because vc-js uses different packages

require('@digitalcredentials/jsonld');
const jsigs = require('@digitalcredentials/jsonld-signatures');

PR #37, opened 2022-06-16
Feature/myc 183/msal authentication

sksadjad:


packages/ms-vc-api-issuer/CHANGELOG.md
```diff
@@ -0,0 +1,36 @@
+# Change Log
```
nklomp, line 1: Please empty the changelog given this is about VC-API issuer

packages/connection-manager/plugin.schema.json
```diff
@@ -0,0 +1,667 @@
+{
```
nklomp, line 1: Why is this part of this PR?

packages/ms-vc-api-issuer/README.md
```diff
@@ -0,0 +1,29 @@
+<!--suppress HtmlDeprecatedAttribute -->
+<h1 align="center">
+  <br>
+  <a href="https://www.sphereon.com"><img src="https://sphereon.com/content/themes/sphereon/assets/img/logo.svg" alt="Sphereon" width="400"></a>
+  <br>Microsoft Azure Active Directory VC API Issuer (Typescript)
```
nklomp, line 5: Please remove "API" from the name everywhere as it is too confusing with VC API, which is an actual W3C specification for a particaly API

packages/ms-vc-api-issuer/__tests__/shared/vcApiIssuerClientCredentialAgentLogic.ts
```diff
@@ -0,0 +1,36 @@
...
+      await testContext.setup()
+      agent = createAgent({
+        plugins:[
+          new MsVcApiIssuer({
+            authenticationType: MsAuthenticationTypeEnum.ClientCredential,
+            authenticationArgs: {
+              azClientId: '<client_id>',
+              azClientSecret:'<client_secret>',
+              azTenantId: '<tenant_id>',
+              credentialManifest:'<credential_manifest>'
```
nklomp, line 20: Hmzz, not sure whether this is going to work in a multitenant setup, as you now need multiple agent instances per manifest and per Azure tenant. I think we might want this as a default for a single tenant scenario, but we need to be able to override the tenant per method invocation (so move it to an options object). The manifest needs to move in all cases. Since a lot of parties will want to issue multiple different credentials

nklomp (changes requested):
Hmzz, I am confused. Is this intended to become the MS Issuer, or is this the authentication library. Right now it seems to be the latter. But as discussed during sprint start (and should also be in the ticket). An auth library should not be an agent, is should simply be a module containing code which can be shared by different future modules, amongst which would be the MS Issuer and Verifier agent modules. So this lib needs to be more like the core module, which isn't an agent. Just a place to put shared code in.

Another important aspect for future issuer/verifier modules is that we might want to add an optional single tenant config on the agent level, but we certainly need to support providing optional (override) tenant options per method invocation or getting an instance of the agent with tenant options, as we need to be able to have a multi-tenant environment.

packages/ms-authenticator/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@sphereon/ms-authenticator",
+  "version": "0.0.1",
```
nklomp, line 3: Please make the version the same as the parent version. This is a multi-module workspace with lerna. Versions are not managed individually

packages/ms-authenticator/src/authenticators/MsAuthenticator.ts
```diff
@@ -0,0 +1,126 @@
...
+ *   credentialManifest: address of your credential manifest. usually in following format:
+ *    https://beta.eu.did.msidentity.com/v1.0/<tenant_id>/verifiableCredential/contracts/<verifiable_credential_schema>
+ * @param authenticationArgs
+ * @constructor
+ */
+export async function ClientCredentialAuthenticator(authenticationArgs: IMsAuthenticationClientCredentialArgs): Promise<string> {
+  var msalConfig = {
+    auth: {
+      clientId: authenticationArgs.azClientId,
+      authority: 'https://login.microsoftonline.com/' + authenticationArgs.azTenantId,
```
nklomp, line 43: Are we sure this always has this form?

packages/ms-authenticator/src/authenticators/MsAuthenticator.ts
```diff
@@ -0,0 +1,126 @@
...
+      loggerOptions: {
+        piiLoggingEnabled: false,
+        logLevel: LogLevel.Verbose,
+      }
+    }
+  }
+
+  const cca = new ConfidentialClientApplication(msalConfig)
+  const msalClientCredentialRequest = {
+    scopes: ['3db474b9-6a0c-4840-96ac-1fceb342124f/.default'],
```
nklomp, line 56: Hmzz, not sure I like this fixed scope in here. Should probably be a param with a default

packages/ms-authenticator/src/authenticators/MsAuthenticator.ts
```diff
@@ -0,0 +1,126 @@
...
+    skipCache: false,
+  }
+  await fetch('https://login.microsoftonline.com/' + authenticationArgs.azTenantId + '/v2.0/.well-known/openid-configuration', {method: 'GET'})
+  .then((res) => res.json())
+  .then(async (resp) => {
+    console.log(`tenant_region_scope = ${resp.tenant_region_scope}`)
```
nklomp, line 62: Remove console.log

packages/ms-authenticator/src/authenticators/MsAuthenticator.ts
```diff
@@ -0,0 +1,126 @@
...
+    let msIdentityHostName = 'https://beta.did.msidentity.com/v1.0/'
+    if (resp.tenant_region_scope == 'EU') {
+      msIdentityHostName = 'https://beta.eu.did.msidentity.com/v1.0/'
```
nklomp, line 65: Are there other regions?

packages/ms-authenticator/src/authenticators/MsAuthenticator.ts
```diff
@@ -0,0 +1,126 @@
...
+    }
+  }
+
+  const cca = new ConfidentialClientApplication(msalConfig)
+  const msalClientCredentialRequest = {
+    scopes: ['3db474b9-6a0c-4840-96ac-1fceb342124f/.default'],
+    skipCache: false,
+  }
+  await fetch('https://login.microsoftonline.com/' + authenticationArgs.azTenantId + '/v2.0/.well-known/openid-configuration', {method: 'GET'})
+  .then((res) => res.json())
```
nklomp, line 60: Are we only doing this to double-check the region. If so I am not sure I like the overhead on every invocation. Then I would rather see a param for the region, against which the authArgs are being checked

packages/ms-authenticator/src/authenticators/MsAuthenticator.ts
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    // get the Access Token
+    try {
+      const result = await cca.acquireTokenByClientCredential(msalClientCredentialRequest)
+      if (result) {
+        return result.accessToken
+      }
+    } catch {
+      console.log('failed to get access token')
```
nklomp, line 79: Please don't swallow exceptions that probably have meaning full errors

packages/ms-authenticator/src/authenticators/MsAuthenticator.ts
```diff
@@ -0,0 +1,126 @@
...
+export async function UsernamePasswordAuthenticator(authenticationArgs: IMsAuthenticationUsernamePasswordArgs): Promise<string> {
+  const msalConfig = {
+    auth: {
+      clientId: authenticationArgs.azClientId,
+      authority: 'https://login.microsoftonline.com/' + authenticationArgs.azTenantId,
+    },
+  }
+  const pca = new PublicClientApplication(msalConfig)
+  return await pca
+  .acquireTokenByUsernamePassword(authenticationArgs as UsernamePasswordRequest)
```
nklomp, line 118: Why is this one shorter than the other one in terms of steps?

packages/ms-authenticator/src/types/IMsAuthenticator.ts
```diff
@@ -0,0 +1,57 @@
...
+    | IMsAuthenticationOnBehalfOfArgs
+}
+
+export interface IMsAuthenticationArgs {
+  azClientId: string
+  azTenantId: string
+}
+export interface IMsAuthenticationClientCredentialArgs extends IMsAuthenticationArgs {
+  azClientSecret: string
+  credentialManifest: string
```
nklomp, line 23: This probably is a URL right?

packages/ms-authenticator/src/types/IMsAuthenticator.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+export interface IMsAuthenticationUsernamePasswordArgs extends IMsAuthenticationArgs {
```
nklomp, line 25: Can we get a bit more documentation on all these params?

packages/ms-authenticator/package.json
sksadjad: changed

packages/ms-authenticator/package.json
```diff
@@ -0,0 +1,49 @@
...
+  "author": "Sphereon <dev@sphereon.com>",
+  "license": "Apache-2.0",
+  "keywords": [
+    "Microsoft",
+    "msal",
+    "Active Directory",
+    "Sphereon",
+    "SSI",
+    "Veramo",
+    "W3C",
```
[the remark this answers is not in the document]
sksadjad: changed

packages/ms-authenticator/src/authenticators/MsAuthenticator.ts
sksadjad: change to read from an optional field, if not, this will be the format

packages/ms-authenticator/src/authenticators/MsAuthenticator.ts
sksadjad: done

packages/ms-authenticator/src/authenticators/MsAuthenticator.ts
```diff
@@ -0,0 +1,126 @@
...
+  .then((response: any) => {
+    console.log('acquired token by password grant', response)
```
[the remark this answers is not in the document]
sksadjad: removed

packages/ms-authenticator/src/authenticators/MsAuthenticator.ts
sksadjad: for the other one `region` is important, therefore we're making a request to determine the region and then go on to authenticate the client

packages/ms-authenticator/src/types/IMsAuthenticator.ts
sksadjad: correct. the code doc specified the format of that url

packages/ms-authenticator/src/types/IMsAuthenticator.ts
sksadjad: added

packages/ms-authenticator/src/authenticators/MsAuthenticator.ts
nklomp: Ping

packages/ms-authenticator/src/authenticators/MsAuthenticator.ts
nklomp: Ping

packages/ms-authenticator/src/authenticators/MsAuthenticator.ts
```diff
@@ -0,0 +1,85 @@
...
+      loggerOptions: {
+        piiLoggingEnabled: false,
+        logLevel: LogLevel.Verbose,
+      }
+    }
+  }
+
+  const cca = new ConfidentialClientApplication(msalConfig)
+  const msalClientCredentialRequest = {
+    scopes: authenticationArgs.scopes ? authenticationArgs.scopes : ['3db474b9-6a0c-4840-96ac-1fceb342124f/.default'],
```
nklomp, line 33: Can we make a const out of this with a meaningful name?

packages/connection-manager/plugin.schema.json
nklomp: Ping

packages/ms-vc-api-issuer/README.md
nklomp: Ping

packages/connection-manager/plugin.schema.json
sksadjad: deleted

packages/ms-vc-api-issuer/README.md
sksadjad: changed

packages/ms-authenticator/src/authenticators/MsAuthenticator.ts
sksadjad: true, but the beta service that is running for the `clientCredential` is dependent on region

packages/ms-authenticator/src/authenticators/MsAuthenticator.ts
sksadjad: changed

packages/ms-authenticator/src/authenticators/MsAuthenticator.ts
```diff
@@ -0,0 +1,85 @@
...
+export async function ClientCredentialAuthenticator(authenticationArgs: IMsAuthenticationClientCredentialArgs): Promise<string> {
+  const msalConfig = {
+    auth: {
+      clientId: authenticationArgs.azClientId,
+      authority: authenticationArgs.authority ? authenticationArgs.authority : 'https://login.microsoftonline.com/' + authenticationArgs.azTenantId,
+      clientSecret: authenticationArgs.azClientSecret,
+    },
+    system: {
+      loggerOptions: {
+        piiLoggingEnabled: false,
```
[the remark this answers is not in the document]
sksadjad: added to optional params, also add optional param for logLevel

PR #38, opened 2022-06-21
MYC-180

BtencateSphereon:
added ion-did-provider


packages/ion-did-provider/src/ion-did-resolver.ts
```diff
@@ -0,0 +1,7 @@
+import * as ION from '@sphereon/ion-tools'
+
+export function IonDIDResolver() {
```
nklomp, line 3: Hmzz, this should probably be more inline with the Resolver package from DIF, conforming to the Interface for resolution. Similar to what we have in the did-uni-client. Veramo also supports that interface for DID resolution

packages/ion-did-provider/__tests__/identity-provider.test.ts
```diff
@@ -0,0 +1,7 @@
+describe('dummy', () => {
+  const a = 100
+  it('should run a dummy test', () => {
```
nklomp, line 3: Ideally we have 2 integration tests. For resolution and registration.

packages/ion-did-provider/src/ion-did-provider.ts
```diff
@@ -0,0 +1,97 @@
...
+    { kms, options }: { kms?: string; options?: any },
+    context: IContext,
+  ): Promise<Omit<IIdentifier, 'provider'>> {
+    const key = await context.agent.keyManagerCreate({ kms: kms || this.defaultKms, type: 'Secp256k1' })
+    // convert to JWK (note, we should have a discussion about key dependencies here)
+    const publicJwk = secp256k1_keyUtils.publicKeyJwkFromPublicKeyHex(key.publicKeyHex)
+
+    const did = new ION.DID({
+      content: {
+        publicKeys: [
```
nklomp, line 33: Questions: Doesn;t it also need 2 other keys like for instance a recoveryKey?

nklomp (changes requested):
Some remarks and a question

packages/ion-did-provider/src/ion-did-provider.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    const identifier: Omit<IIdentifier, 'provider'> = {
+      did: await did.getURI(),
+      controllerKeyId: key.kid,
+      keys: [key],
+      services: [],
+    }
+
+    // TODO: Long version DID is not conformant to "initial value" definition.
+    console.log(`Created (short version): ${ await did.getURI('short') }`)
```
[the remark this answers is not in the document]
BtencateSphereon: changed to debug

packages/ion-did-provider/src/ion-did-resolver.ts
BtencateSphereon: did some refactoring on the ion resolver

packages/ion-did-provider/src/ion-did-provider.ts
BtencateSphereon: As discussed, created the following two tickets. Hopefully I wrote it down correctly

https://sphereon.atlassian.net/browse/WAL-204
https://sphereon.atlassian.net/browse/WAL-206

packages/ion-did-provider/src/ion-did-resolver.ts
nklomp: LGTM

nklomp (approved):
LGTM

PR #49, opened 2022-08-02
Feature/wal 222

nklomp:
Implement migrations. Also had to update most entities from the data-store to make them compatible with both sqlite and postgres.

Draft, since the mnemonic seed module needs similar work


nklomp:
Closed and opened https://github.com/Sphereon-Opensource/ssi-sdk/pull/51 instead

PR #51, opened 2022-08-03
Add migration support

nklomp:
Implement migrations. Also had to update most entities from the data-store to make them compatible with both sqlite and postgres.

Draft, since the mnemonic seed module needs similar work


nklomp:
Ready for review:

Please note that current version is ubstable.70, which is lower than an already published .92. Has to do with CI/CD being confused by a branch being replaced by me. Any code depending on this PR will need an update to a -next version after this has been merged to develop

packages/data-store-common/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright 2019 Consensys AG
```
BtencateSphereon, line 189: nitpicking here, should this not be 2022 and Sphereon BV?

packages/data-store-common/src/__tests__/entities.test.ts
```diff
@@ -0,0 +1,145 @@
...
+    dbConnection = await createConnection({
+      type: 'sqlite',
+      database: ':memory:',
+      // logging: 'all',
+      migrationsRun: false,
+      migrations: DataStoreMigrations,
+      synchronize: false,
+      entities: DataStoreConnectionEntities,
+    })
+    // await dbConnection.dropDatabase()
```
BtencateSphereon, line 22: this can be removed?

packages/data-store-common/LICENSE
nklomp: Hehe, guess someone copied some stuff over :D

packages/data-store-common/src/__tests__/entities.test.ts
nklomp: Yeah will remove the dropDatabase. If you don't mind I will leave the logging comment. This allows someone to enable it quickly in case of problems with the test

nklomp:
Addressed the comments. Also ensured we actually apply the migrations to our internal reference DBs. Added a bit of documentation about that as well in the READMEs.

BtencateSphereon (approved):
LGTM

PR #53, opened 2022-08-10
Feature/wal 246 ION DID Provider

nklomp:
Implement ION DID Provider. Seeking initial feedback before rebasing to Veramo and opening a PR there.

Changes:
- Remove old ION provider code
- Do not depend on ion-tools
- Do not depend on external POW SDK. An improved version is part of the code. We might extract that into a seperate lib at one point
- Update and recovery keys, including key rotations
- Full implementation of all Identifier methods

Known issues:
- POW works on Node. Not tested yet on React-native. Out of scope for this implementation
- Adding/Updating services/keys does not get propagated. Needs more investigation. MS gives back a 200 response for them, but it never ends up in the DID Document. Will look at that later.


packages/ion-did-provider/README.md
```diff
@@ -1,17 +1,260 @@
...
+from the ION
+node, right now these update and delete operations are not reflected in the eventual DID Document after the anchor time
+has passed. We
+are actively looking to solve this issue_
 
-```shell
-yarn add @sphereon/react-native-argon2
+## Long form and short form ION DIDs
+
+ION has the concept of short form DIDs. a Short form DID is a DID, you also typically encounter when using other DID
```
BtencateSphereon, line 18: A Short form DID is a DID, you also typically encounter when using other DIDs ?

BtencateSphereon (approved):
LGTM

PR #54, opened 2022-08-10
VDX-88

BtencateSphereon:


BtencateSphereon:
plugin.schema.json is supplied as we cannot let it generate one as callbacks are being used

BtencateSphereon:
the build failing is because of a test in the ms-authenticator that takes to long to complete

packages/wellknown-did-verifier/README.md
```diff
@@ -0,0 +1,95 @@
...
+    new WellKnownDidVerifier({
+      signatureVerifications: {'verified': () => Promise.resolve({ verified: true })},
+      onlyVerifyServiceDids: true
+    }),
+  ],
+})
+```
+
+### Register signature verification callback:
+
```
nklomp, line 44: Could we add a line what this is actually doing?

packages/wellknown-did-verifier/README.md
```diff
@@ -0,0 +1,95 @@
...
+agent.registerSignatureVerification({
+  key: 'example_key',
+  signatureVerification: () => Promise.resolve({ verified: true })
+})
+  .then(() => console.log('success'))
+  .catch(() => console.log('failed'))
+```
+
+### Remove signature verification callback:
+
```
nklomp, line 55: Add a line what this is doing

packages/wellknown-did-verifier/README.md
```diff
@@ -0,0 +1,95 @@
...
+```typescript
+agent.removeSignatureVerification({ key: 'example_key' })
+  .then(() => console.log('success'))
+  .catch(() => console.log('failed'))
+```
+
+### Verify domain linkage:
+
```
nklomp, line 63: Add a line what this is doing

packages/wellknown-did-verifier/__tests__/localAgent.test.ts
```diff
@@ -0,0 +1,76 @@
...
+  await localAgent.registerSignatureVerification({
+    key: 'verified',
+    signatureVerification: () => Promise.resolve({ verified: true })
+  }, null)
+
+  const DID = 'did:key:z6MkoTHsgNNrby8JzCNQ1iRLyW5QQ6R8Xuu6AA8igGrMVPUM';
+  const ORIGIN = 'https://example.com';
+  const DOCUMENT = {
+    '@context': [
+        'https://www.w3.org/ns/did/v1',
```
nklomp, line 23: Doesn't look like it is formatted with prettier

packages/wellknown-did-verifier/__tests__/restAgent.test.ts
```diff
@@ -0,0 +1,114 @@
...
+      },
+      {
+        id: `${DID}#bar`,
+        type: ServiceTypesEnum.LINKED_DOMAINS,
+        serviceEndpoint: ORIGIN,
+      },
+    ],
+  };
+
+  agent.resolveDid = jest.fn().mockReturnValue(Promise.resolve({ didDocument: DOCUMENT }))
```
nklomp, line 79: Isn't this exactly the same as in the setup() method? (The Document and mock)

packages/wellknown-did-verifier/__tests__/shared/wellKnownDidVerifierAgentLogic.ts
```diff
@@ -0,0 +1,171 @@
...
+
+    beforeAll(async () => {
+      await testContext.setup()
+      agent = testContext.getAgent()
+    })
+
+    afterAll(testContext.tearDown)
+
+    it('should verify domain linkage with signature verification key', async () => {
+      nock('https://example.com').get('/.well-known/did-configuration.json').times(3).reply(200, DID_CONFIGURATION);
```
nklomp, line 35: Is the 3 times there because we have 3 times a link in there, or is there another reason?

packages/wellknown-did-verifier/agent.yml
```diff
@@ -0,0 +1,107 @@
+version: 3.0
+
+constants:
+  baseUrl: http://localhost:3335
+  port: 3335
+  # please use your own X25519 key, this is only an example
+  secretKey: 29739248cad1bd1a0fc4d9b75cd4d2990de535baf5caadfdf8d8f86664aa830c
```
nklomp, line 7: What is the secretKey being used for?

packages/wellknown-did-verifier/package.json
```diff
@@ -0,0 +1,43 @@
+{
+  "name": "@sphereon/ssi-sdk-wellknown-did-verifier",
+  "version": "0.6.0",
```
nklomp, line 3: develop is at 0.7.x. So a change to 0.7.0 should be enough

packages/wellknown-did-verifier/src/types/IWellKnownDidVerifier.ts
```diff
@@ -0,0 +1,48 @@
...
+  key: string
+  signatureVerification: (args: IVerifyCallbackArgs) => Promise<IVerifyCredentialResult>
+}
+
+export interface IRemoveSignatureVerificationArgs {
+  key: string
+}
+
+export interface IVerifyDomainLinkageArgs {
+  didUrl: string
```
nklomp, line 36: Is there a reason why this isn't called a did? It really expects a DID there instead of a didUrl right?

packages/wellknown-did-verifier/src/types/IWellKnownDidVerifier.ts
```diff
@@ -0,0 +1,48 @@
...
+  verifyDidConfigurationResource(args: IVerifyDidConfigurationResourceArgs, context: IRequiredContext): Promise<IResourceValidation>
+}
+
+export interface IWellKnownDidVerifierOptionsArgs {
+  signatureVerifications?: Record<string, (args: IVerifyCallbackArgs) => Promise<IVerifyCredentialResult>>
+  onlyVerifyServiceDids?: boolean
+}
+
+export interface IRegisterSignatureVerificationArgs {
+  key: string
```
nklomp, line 27: Can we change the name. This might be confusing with did private/public keys

nklomp (changes requested):
Basically some small remarks about some documentation and naming

packages/wellknown-did-verifier/README.md
BtencateSphereon: added description

packages/wellknown-did-verifier/README.md
BtencateSphereon: added description

packages/wellknown-did-verifier/README.md
BtencateSphereon: added description

packages/wellknown-did-verifier/__tests__/localAgent.test.ts
BtencateSphereon: fixed

packages/wellknown-did-verifier/__tests__/restAgent.test.ts
BtencateSphereon: not sure if i understand the question. so the reason why this is a bity weird is because i had to mock the context of the agent. i tried mocking it directly in the tests but that did not work. so had to duplicate it for both local and rest tests

packages/wellknown-did-verifier/__tests__/shared/wellKnownDidVerifierAgentLogic.ts
BtencateSphereon: we have 3 origins in the did document yes

packages/wellknown-did-verifier/agent.yml
BtencateSphereon: got copied over, removed it now

packages/wellknown-did-verifier/package.json
BtencateSphereon: fixed

packages/wellknown-did-verifier/__tests__/restAgent.test.ts
nklomp: Yeah I meant the setup() method in the localAgent. Not a biggie

packages/wellknown-did-verifier/src/types/IWellKnownDidVerifier.ts
BtencateSphereon: The thought here was that since we need to resolve the DID first and the resolver can have a didUrl. Then from the DID document the only DID's that can be passed on are the ones in the service. For the verifyResource function it is stated as a did

packages/wellknown-did-verifier/__tests__/shared/wellKnownDidVerifierAgentLogic.ts
nklomp: Just wanted to make sure we do not hit the endpoint multiple times during validation because that would be sub-optimal

packages/wellknown-did-verifier/src/types/IWellKnownDidVerifier.ts
BtencateSphereon: renamed

packages/wellknown-did-verifier/src/types/IWellKnownDidVerifier.ts
```diff
@@ -24,12 +20,12 @@ export interface IWellKnownDidVerifierOptionsArgs {
 }
 
 export interface IRegisterSignatureVerificationArgs {
-  key: string
+  signatureVerificationKey: string
```
nklomp, line 23: callbackName might be better, because this can still be confused with a public key for signature verification

packages/wellknown-did-verifier/src/types/IWellKnownDidVerifier.ts
```diff
@@ -24,12 +20,12 @@ export interface IWellKnownDidVerifierOptionsArgs {
...
   signatureVerification: (args: IVerifyCallbackArgs) => Promise<IVerifyCredentialResult>
 }
 
 export interface IRemoveSignatureVerificationArgs {
-  key: string
+  signatureVerificationKey: string
```
nklomp, line 28: callbackName might be better, because this can still be confused with a public key for signature verification

packages/wellknown-did-verifier/src/types/IWellKnownDidVerifier.ts
nklomp: A did resolver can have a DID url because it can actually resolve subsections of a DID or in some cases even external resources present in a DID document. In here we only should have a did, not a didUrl. So it really should be did

packages/wellknown-did-verifier/src/types/IWellKnownDidVerifier.ts
nklomp: See other remarks. It can still be consufed with a public key for signature verification. Suggest to name it to something like callbackName, which makes immediate sense to a developer

BtencateSphereon: renamed

packages/wellknown-did-verifier/src/types/IWellKnownDidVerifier.ts
BtencateSphereon: renamed

nklomp (approved):
LGTM

PR #55, opened 2022-08-22
Feature/vdx 123/bls key resolver

maikel-maas:


packages/tsconfig.json
```diff
@@ -14,6 +14,7 @@
     { "path": "waci-pex-qr-react" },
     { "path": "bls-kms-local" },
     { "path": "bls-did-provider-key" },
+    { "path": "bls-key-did-resolver" },
```
nklomp, line 17: Can we rename it to bls-did-resolver-key, more inline with the providers?

maikel-maas: Done!

packages/vc-handler-ld-local/README.md
```diff
@@ -49,3 +49,114 @@ yarn test
...
 - `yarn fix` - runs `eslint --fix` as well as `prettier` to fix code style.
+
+### Request examples:
+
+#### Issue Verifiable Credential
+
+Ed25519:
+
+```shell
+POST http://localhost:7071/api/vdx-ivcs/v1/createVerifiableCredentialLDLocal
```
nklomp, line 60: This needs more context. This is using Veramo as a REST API, whilst most will not use it that way. So either make that known, or change the examples to just use the agent directly.

nklomp (changes requested):
Some remarks.

I think we need to make a comment in the README why we created this module. What was missing from Veramo. Because probably we will want open a PR upstream to get it in Veramo itself

packages/bls-did-resolver-key/src/ed25519.ts
```diff
@@ -0,0 +1,41 @@
...
+  const bytes = new Uint8Array(key.length + 2)
+  bytes[0] = 0xec
+  bytes[1] = 0x01
+  bytes.set(key, 2)
+  return `z${u8a.toString(bytes, 'base58btc')}`
+}
+export const keyToDidDoc = (pubKeyBytes: Uint8Array, fingerprint: string): DIDDocument => {
+  const did = `did:key:${fingerprint}`
+  const keyId = `${did}#${fingerprint}`
+  const x25519PubBytes = ed25519.convertPublicKeyToX25519(pubKeyBytes)
```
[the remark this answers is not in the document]
maikel-maas: The code comes from the original plugin.

packages/vc-handler-ld-local/README.md
maikel-maas: Fixed!

packages/bls-did-resolver-key/src/ed25519.ts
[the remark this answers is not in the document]
nklomp: It is just odd to be mixing them in the same code. But hey, we have more important things to worry about.

packages/vc-handler-ld-local/README.md
```diff
@@ -50,7 +50,165 @@ There are other utility scripts that help with development.
...
+          store: new DIDStore(connection),
+          defaultProvider: 'did:key',
+          providers: {
+            'did:key': new BlsKeyDidProvider({
+              defaultKms: 'local',
+            }),
+            'did:web': new WebDIDProvider({
+              defaultKms: 'local',
+            }),
+            'did:factom': new FactomDIDProvider({
```
nklomp, line 78: Please remove the providers not being used in the example. Main reason is that I don't want our internal registrar links to end up all over our repos until VDX is in production

nklomp (changes requested):
Single small fix needed for examples

packages/vc-handler-ld-local/README.md
maikel-maas: Removed

nklomp (approved):

PR #61, opened 2022-09-15
VDX-135 Verify the VC signatures of linked domain

maikel-maas:
- Made id optional in the credential;

### DIFF
--- a/packages/ssi-types/src/types/vc.ts
+++ b/packages/ssi-types/src/types/vc.ts
@@ -11,7 +11,7 @@ export interface ICredential {
   // If sub is present, the value MUST be used to set the value of the id property of credentialSubject of the new credential JSON object.
   credentialSubject: ICredentialSubject
   // If jti is present, the value MUST be used to set the value of the id property of the new JSON object.
-  id: string
+  id?: string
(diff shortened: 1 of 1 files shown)


packages/ssi-types/src/types/vc.ts
```diff
@@ -76,6 +76,7 @@ export interface IHasProof {
 export type IVerifiableCredential = ICredential & IHasProof
 
 export interface IPresentation {
+  id?: string
```
sksadjad, line 79: https://www.w3.org/TR/vc-data-model/#presentations-0

packages/ssi-types/src/types/vc.ts
```diff
@@ -11,7 +11,7 @@ export interface ICredential {
   // If sub is present, the value MUST be used to set the value of the id property of credentialSubject of the new credential JSON object.
   credentialSubject: ICredentialSubject
   // If jti is present, the value MUST be used to set the value of the id property of the new JSON object.
-  id: string
+  id?: string
```
sksadjad, line 14: https://www.w3.org/TR/vc-data-model/#identifiers

sksadjad (approved):
approving this based on the two links that mentioned in the comments

PR #65, opened 2022-10-04
Allow multiple subjects and fix Credential expiration date location

nklomp:
- Allow to have multiple subjects in a Credential, according to VC Data Model spec.
- Fix bug where expirationDate was made part of the subject, whilst it is part of the credential metadata itself


sksadjad (approved):
looks good to me

PR #66, opened 2022-10-12
feature: added VC and VP format

sksadjad:


packages/ssi-types/src/types/vc.ts
```diff
@@ -195,6 +195,8 @@ export interface JwtDecodedVerifiablePresentation {
 }
 
 export type ClaimFormat = 'jwt' | 'jwt_vc' | 'jwt_vp' | 'ldp' | 'ldp_vc' | 'ldp_vp' | string
+export type CredentialFormat = 'jwt' | 'ldp' | string
```
nklomp, line 198:
```suggestion
export type CredentialFormat = 'jwt' | 'ldp' | 'jwt_vc' | 'ldp_vc' | string
```

packages/ssi-types/src/types/vc.ts
```diff
@@ -195,6 +195,8 @@ export interface JwtDecodedVerifiablePresentation {
 }
 
 export type ClaimFormat = 'jwt' | 'jwt_vc' | 'jwt_vp' | 'ldp' | 'ldp_vc' | 'ldp_vp' | string
```
nklomp, line 197:
```suggestion
export type ClaimFormat = CredentialFormat | ClaimFormat
```

nklomp: Needs to be moved below the other properties probably

nklomp (changes requested):
See suggestions

packages/ssi-types/src/types/vc.ts
sksadjad: moved claim lower than `CredentialFormat` and `PresentationFormat`

nklomp (approved):
LGTM

PR #68, opened 2022-10-17
JFF-1

BtencateSphereon:


packages/jwk-did-provider/__tests__/jwk-did-resolver.test.ts
```diff
@@ -0,0 +1,99 @@
...
+      store: new MemoryDIDStore(),
+    }),
+  ],
+})
+
+describe('@sphereon/jwk-did-resolver', () => {
+  it('should resolve did:jwk', async () => {
+    const didResolutionResult: DIDResolutionResult = await agent.resolveDid({ didUrl: 'did:jwk:eyJrdHkiOiJFQyIsImNydiI6InNlY3AyNTZrMSIsIngiOiJCS2NlRjMwbHBTNkptT1RsS09LQVdudGtKdVRCSzNGX1JoaXlEcTRtdm9jIiwieSI6Im9WY1phQnpiSFJ2UW5iSXhwRWRXbVlRMGtSRm42ajVDRkVQcGxvX09ON1UifQ' })
+
+    expect(didResolutionResult.didResolutionMetadata).toBeDefined()
```
nklomp, line 40: Could you check against exact values in this class? 
We then get to catch any regressions.

packages/jwk-did-provider/package.json
```diff
@@ -9,6 +9,7 @@
   },
   "dependencies": {
     "@ethersproject/random": "^5.6.1",
+    "@sphereon/ssi-types": "^0.8.0",
```
nklomp, line 12: Is there also a reference value in tsConfig?

nklomp (changes requested):
Some small remarks.

I think the tests should match the exact JSON Documents, so we can catch any regressions

packages/jwk-did-provider/package.json
BtencateSphereon: added

packages/jwk-did-provider/__tests__/jwk-did-resolver.test.ts
BtencateSphereon: changed

nklomp (approved):
LGTM

PR #71, opened 2022-11-03
bugfix/WAL-223/pex

maikel-maas:


packages/did-auth-siop-op-authenticator/package.json
```diff
@@ -13,8 +13,8 @@
     "build": "tsc --build"
   },
   "dependencies": {
-    "@sphereon/did-auth-siop": "0.2.14",
-    "@sphereon/pex": "1.1.4",
+    "@sphereon/did-auth-siop": "0.2.15-unstable.0",
+    "@sphereon/pex": "2.0.0-unstable.3",
```
nklomp, line 17: Do not update to 2.0.0 untill we have actually released that. They are not API compatible

PR #73, opened 2022-11-14
Feature/ssi types/add syntactic sugar

sksadjad:
reading Gaia-X docs, I've noticed a slight difference between our types and Gaia-X examples. So I went back to vc-data-model and found [this](https://www.w3.org/TR/vc-data-model/#syntactic-sugar) section.
This PR covers the first bullet point of the section above.

### DIFF
--- a/packages/ssi-types/src/mapper/credential-mapper.ts
+++ b/packages/ssi-types/src/mapper/credential-mapper.ts
@@ -66,6 +66,7 @@ export class CredentialMapper {
       presentation: {
         ...vp,
         verifiableCredential: vcs, // We overwrite the credentials with wrapped versions, making it an InternalVerifiablePresentation. Note: we keep the singular key name of the vc data model
+        type: vp.type? vp.type: vp['@type']
       },
       vcs,
     }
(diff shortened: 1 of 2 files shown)


nklomp (changes requested):
Not going to approve. Do you have a ref to their examples?
The implementation must be wrong. The syntactic sugar is explaining that what would be described in pure JSON-LD is aliased to type and id respectively and thus should be used.

If we however would need something like this I rather have us approach this in a uniform way.

So creating separate interfaces out of it. Then when they are required using an intersection and union like done for ICredential. When they are optional using an intersection and a union of 2 Partial.

Reason is that these @id vs ids, are basically everywhere. Your suggestion would not be the only places.

nklomp:
> @nklomp it's from gaia-x gitlab. https://gitlab.com/gaia-x/technical-committee/federation-services/self-description-model/-/blob/main/docs/intro.md#example
> 
> and the section of vc-data-model that I referenced here is just talking about two properties: `id` and `type`

Will have a look. It talks about 3 constructs from json-ld which are present at all object layers/contexts. By extension that would have to mean it should be applicable to all for consistency. A quick look at the example also already shows that

sksadjad:
based on your remarks, I've closed this pr

PR #79, opened 2023-01-17
WAL-380

BtencateSphereon:


packages/connection-manager/__tests__/shared/connectionManagerAgentLogic.ts
```diff
@@ -81,6 +81,19 @@ export default (testContext: { getAgent: () => ConfiguredAgent; setup: () => Pro
       expect(result.length).toBeGreaterThan(0)
     })
 
+    it('should get parties by filter', async () => {
```
nklomp, line 84: Can we also have some testcases, which only have one property (like only name), but more importantly have a test that shouldn't return anything.

nklomp (changes requested):
Small request for testcases. Rest LGTM

packages/connection-manager/__tests__/shared/connectionManagerAgentLogic.ts
BtencateSphereon: added additional test cases

nklomp (approved):

nklomp:
LGTM

nklomp (approved):
LGTM

PR #84, opened 2023-03-03
ProofPurpose types

sanderPostma:


nklomp (changes requested):
Why are there so many classes in what should be interfaces probably?  Why all of the requires instead of imports? We need to move to ES modules, so why use requires then?

sanderPostma:
This is still a work in progress. I have no idea what I am doing yet, still experimenting how typescript works. I have to push the branch to see how the tests run because locally I get so many errors they don't fit the console buffer.

nklomp:
All that is needed is creating interfaces for the ProofPurposes. The functions/classes to use these are already available. We simply cannot expose them as the underlying project isn't a TS project. That is why the current typeof constructs give errors in ES6.

nklomp:
What you could consider is creating classes to delegate it to the original lib, to make them more easily available for people using the SSI-SDK

sanderPostma:
> Why are there so many classes in what should be interfaces probably? Why all of the requires instead of imports? We need to move to ES modules, so why use requires then?
> 
> Also exposing classes in interfaces should be avoided

You are talking about the classes in exttypes/purposes. I did not make those, it came from a command
`npx -p typescript tsc src/**/*.js --declaration --allowJs --emitDeclarationOnly --outDir types`
I assumed that is valid typescript. I am still struggling with the constructors containing these inline {objects, like, this }

> What you could consider is creating classes to delegate it to the original lib, to make them more easily available for people using the SSI-SDK
Will look into it

PR #85, opened 2023-03-06
Feature/vdx 173 schema gen

sanderPostma:


packages/vc-handler-ld-local/src/types/types.ts
```diff
@@ -218,14 +218,25 @@ export interface IVerifyPresentationLDArgs {
...
-  presentationPurpose?: typeof ProofPurpose
+  presentationPurpose?: IProofPurpose
 
   /**
    * Check status function, to check credentials that have a credentialStatus property
    */
   checkStatus?: Function
 }
 
+export interface IProofPurpose {
```
sanderPostma, line 229: I could have made different types and or them like AuthenticationProofPurpose|  ControllerProofPurpose | AssertionProofPurpose, but I'm not so confident the API extractor will turn this into a correct OpenAPI v3 structure.

sanderPostma:
Still not building consistently

packages/vc-handler-ld-local/src/types/types.ts
nklomp: Please do try that first though. The Proof Purposes are well known constructs in any SSI solution. This is really confusing to people that know SSI.

nklomp (changes requested):
See remark. We really need different proof purposes, as these are core parts of any SSI solution. Current approach will confuse people that know SSI.
Also I would not see why it would go wrong, as anyofs  / inheritence is fully supported.

PR #89, opened 2023-03-09
Feature/vdx 174 velocity message handler

sanderPostma:


packages/ssi-types/src/mapper/credential-mapper.ts
```diff
@@ -217,13 +227,19 @@ export class CredentialMapper {
...
+    }
+
     uniformPresentation.verifiableCredential = uniformPresentation.verifiableCredential?.map((vc) =>
       CredentialMapper.toUniformCredential(vc, opts)
     ) as IVerifiableCredential[] // We cast it because we IPresentation needs a VC. The internal Credential doesn't have the required Proof anymore (that is intended)
     return uniformPresentation
   }
 
-  static jwtEncodedCredentialToUniformCredential(jwt: string, opts?: { maxTimeSkewInMS?: number }): IVerifiableCredential {
+  static jwtEncodedCredentialToUniformCredential(jwt: string, opts?: { maxTimeSkewInMS?: number }): ICredential {
```
sanderPostma, line 242: @nklomp not sure of I can do this here. The credential provided by Velocity does not have a proof section.

nklomp: No don't. The code adds the JWT as proof when it converts/decodes it to the Uniform VC (or it should do that)

packages/ssi-types/src/mapper/credential-mapper.ts
```diff
@@ -217,13 +227,19 @@ export class CredentialMapper {
       isJwtEncoded || isJwtDecoded
         ? CredentialMapper.jwtDecodedPresentationToUniformPresentation(decoded as JwtDecodedVerifiablePresentation, false)
         : (decoded as IVerifiablePresentation)
+
+    // At time of writing Velocity Networks does not conform to specification. Adding bare minimum @context section to stop parsers from crashing and whatnot
+    if (!uniformPresentation['@context']) {
```
nklomp, line 231: Please make this an option. I don't want us to blindly do this, as this is non-spec compliant

packages/vp-message-handler/src/message-handler.ts
```diff
@@ -0,0 +1,35 @@
...
+ * A plugin for {@link @sphereon/ssi-sdk-vp-message-handler#VpMessageHandler} that unpacks & uniforms a VC/VP
+ * @public
+ */
+export class VpMessageHandler extends AbstractMessageHandler {
+  async handle(message: Message, context: IContext): Promise<Message> {
+    if (message.raw) {
+      try {
+        const parsed = JSON.parse(message.raw)
+        if (parsed) {
+          if (parsed['jwt_vp']) {
```
nklomp, line 20: replace if if with &&

packages/vp-message-handler/src/message-handler.ts
```diff
@@ -0,0 +1,35 @@
...
+            message.raw = parsed['jwt_vp']
+            const jwtDecodedVp = CredentialMapper.toWrappedVerifiablePresentation(message.raw as OriginalVerifiablePresentation).decoded
+            if (jwtDecodedVp.aud) {
+              message.addMetaData({ type: 'JWT' })
```
nklomp, line 24: Is there are reason why we are not passing the wrapped VP in the message from this point on? We typically need to know the content if the VCs/VPs anyway

nklomp (changes requested):
See remarks.

Adding the context should only happen when an optional argument is set, as it is non spec compliant and I want to stay away from too much non-spec manipulation happening automatically. In case we are in a velocity environment it means we need to make sure to enable the option somehow. So either by making sure we call the function with the respective argument, or by creating something that resembles a configuration option

packages/vp-message-handler/src/message-handler.ts
sanderPostma: At this point I am only unwrapping the raw credential from the message. toWrappedVerifiablePresentation is only done to mark it as JWT credential in the Veramo message.

The SharePoint agent module will decode it again to get the field info for the mappings. The original JWT will by the file data as it still contains the signature.

I can put the decoded version as message.data, it will save double work, I will try taht now. The "toUniformPresentation" call I would like to keep in the SharePoint module because now I need to call it with new option "repairMissingVpContext"

packages/ssi-types/src/mapper/credential-mapper.ts
sanderPostma: Added optional repairMissingVpContext option

packages/vp-message-handler/src/message-handler.ts
sanderPostma: done

packages/ssi-types/src/mapper/credential-mapper.ts
```diff
@@ -217,6 +229,12 @@ export class CredentialMapper {
       isJwtEncoded || isJwtDecoded
         ? CredentialMapper.jwtDecodedPresentationToUniformPresentation(decoded as JwtDecodedVerifiablePresentation, false)
         : (decoded as IVerifiablePresentation)
+
+    // At time of writing Velocity Networks does not conform to specification. Adding bare minimum @context section to stop parsers from crashing and whatnot
+    if (opts?.repairMissingVpContext && !uniformPresentation['@context']) {
```
nklomp, line 234: addContextIfMissing

packages/vp-message-handler/package.json
```diff
@@ -0,0 +1,42 @@
...
+  "version": "0.8.0",
+  "source": "src/index.ts",
+  "main": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "scripts": {
+    "build": "tsc --build",
+    "generate-plugin-schema": "yarn veramo dev extract-api"
+  },
+  "dependencies": {
+    "@veramo/core": "^4.2.0",
```
nklomp, line 12: Please fixate everything on 4.2.0 exactly

nklomp (changes requested):
Some small remarks

PR #92, opened 2023-03-15
Bugfix/vdx 161

maikel-maas:


nklomp:
Please rebase on develop @maikel-maas

package.json
```diff
@@ -68,8 +68,9 @@
     "@veramo/remote-server": "4.2.0",
     "@veramo/selective-disclosure": "4.2.0",
     "@veramo/url-handler": "4.2.0",
+    "typeorm": "0.3.10",
```
nklomp, line 71: If you are doing a global resolution, why are you changing the the values in the subpackages, that are not really changing much of the behaviour. Basically Veramo has ^0.3.10. Without this resolution the end result of any implementation still is  0.3.12, given that version satisfies both the original Veramo version range as well as ours.

In other words the other changes do nothing

packages/mnemonic-seed-manager/__tests__/localAgent.test.ts
```diff
@@ -15,14 +13,13 @@ import { OrPromise } from '@veramo/utils'
...
 const KMS_SECRET_KEY = 'd17c8674f5db9396f8eecccde25e882bb0336316bc411ae38dc1f3dcd7ed100f'
-let databaseFile = 'database.sqlite'
 let dbConnection: OrPromise<DataSource>
 let agent: any
 
 const setup = async (): Promise<boolean> => {
   const db: OrPromise<DataSource> = new DataSource({
     type: 'sqlite',
-    database: databaseFile,
+    database: ':memory:',
```
nklomp, line 22: Not sure whether I like this change, as it could just be masquerading stuff. Why is this part of the fix?

nklomp (commented):
If you ask me the only thing from this review which has an impact is the addition of the awaits and the global resolution. The rest should not be changed

packages/mnemonic-seed-manager/__tests__/localAgent.test.ts
maikel-maas: We have packages using an in-memory database and other ones using a file. I've changed it for the sake of consistency, to avoid to have a database file in the repository and to spare some milliseconds running the tests.

package.json
maikel-maas: This is the code from the root package.json:

```json
"resolutions": {
    "@veramo/cli": "4.2.0",
    "@veramo/core": "4.2.0",
    "@veramo/credential-w3c": "4.2.0",
    "@veramo/data-store": "4.2.0",
    "@veramo/did-jwt": "4.2.0",
    "@veramo/credential-ld": "4.2.0",
    "@veramo/utils": "4.2.0",
    "@veramo/did-manager": "4.2.0",
    "@veramo/kms-local": "4.2.0",
    "@veramo/key-manager": "4.2.0",
    "@veramo/credential-eip712": "4.2.0",
    "@veramo/did-comm": "4.2.0",
    "@veramo/message-handler": "4.2.0",
    "@veramo/did-discovery": "4.2.0",
    "@veramo/did-provider-ethr": "4.2.0",
    "@veramo/did-provider-key": "4.2.0",
    "@veramo/did-provider-web": "4.2.0",
    "@veramo/did-resolver": "4.2.0",
    "@veramo/remote-client": "4.2.0",
    "@veramo/remote-server": "4.2.0",
    "@veramo/selective-disclosure": "4.2.0",
    "@veramo/url-handler": "4.2.0",
    "typeorm": "0.3.10",
    "**/@mattrglobal/jsonld-signatures-bbs/jsonld": "4.0.1",
    "**/jsonld": "link:./node_modules/@digitalcredentials/jsonld",
    "**/isomorphic-webcrypto": "link:./node_modules/@sphereon/isomorphic-webcrypto"
  }
```
as you can see  *"typeorm": "0.3.10"* is already included in the resolutions

packages/mnemonic-seed-manager/__tests__/localAgent.test.ts
nklomp: DB file should just be in .gitignore if you worry about that.

package.json
nklomp: Okay then the resolution value is wrong or something else is up.

Veramo is using ^0.3.10 and we were using ^0.3.12. Since 0.3.12 is the latest it should resolve to 0.3.12 for both.

In other words, you should find out which package apparently is causing a hard fix on another version, as we are now just removing symptomps without knowing the actual cause

nklomp (commented):
See remarks. We still have not found the cause, and that is what we really should be doing. This is just a workaround

nklomp:
Closing this one, as this is highly unlikely to solve any bug as ^0.3.10 and ^0.3.12 both result in 0.3.15 currently. Changing the 0.3.12 to 0.3.10 wouldn't solve that

PR #94, opened 2023-03-20
feature/VDX-176/openid4vci_qr

maikel-maas:
… qr-code-generator plugin

   - Created QR code for Openid4VCI;
   - Ran prettier;

### DIFF
--- a/packages/qr-code-generator/src/agent/QrCodeProvider.tsx
+++ b/packages/qr-code-generator/src/agent/QrCodeProvider.tsx
@@ -6,6 +6,7 @@ import {
   DIDCommV2OOBInvitationData,
   IQRCodeGenerator,
   IRequiredContext,
+  OpenID4VCIDataWithScheme,
   QRType,
   SIOPv2DataWithScheme,
   URIData,
@@ -23,6 +24,8 @@ export class QrCodeProvider implements IAgentPlugin {
     qrSIOPv2Element: QrCodeProvider.qrSIOPv2Element.bind(this),
     qrSIOPv2Value: QrCodeProvider.qrSIOPv2Value.bind(this),
     qrURIElement: QrCodeProvider.qrURIElement.bind(this),
+    qrOpenID4VCIElement: QrCodeProvider.qrOpenID4VCIElement.bind(this),
+    qrOpenID4VCIValue: QrCodeProvider.qrOpenID4VCIValue.bind(this),
   }
 
   /** {@inheritDoc IQRCodeGenerator.uriElement} */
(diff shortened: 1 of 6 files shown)


packages/qr-code-generator/src/types/IQRCodeGenerator.ts
```diff
@@ -36,11 +39,10 @@ export interface ValueResult<T extends QRType, D> {
 }
 
 export enum QRType {
-  // OIDC4VCI = 'openid-credential-offer',
-
   URI = 'uri',
   SIOPV2 = 'openid-vc',
   DIDCOMM_V2_OOB_INVITATION = 'https://didcomm.org/out-of-band/2.0/invitation',
+  OIDC4VCI = 'openid-credential-offer',
```
nklomp, line 45: OpenID4VCI please

packages/qr-code-generator/src/types/IQRCodeGenerator.ts
```diff
@@ -49,6 +51,15 @@ export interface SIOPv2DataWithScheme {
   requestUri: string
 }
 
+export type OpenID4VCIScheme = 'openid-credential-offer' | 'https' | string
+export interface OpenID4VCIDataWithScheme {
+  scheme?: OpenID4VCIScheme
```
nklomp, line 56: Please replace domain and path with an optional baseUri

packages/qr-code-generator/src/agent/QrCodeProvider.tsx
```diff
@@ -71,12 +74,46 @@ export class QrCodeProvider implements IAgentPlugin {
...
+    )
+  }
+
+  /** {@inheritDoc IQRCodeGenerator.qrOpenID4VCIValue} */
+  private static async qrOpenID4VCIValue(
+    args: CreateValueArgs<QRType.OIDC4VCI, OpenID4VCIDataWithScheme>,
+    context: IRequiredContext
+  ): Promise<string> {
+    const { object } = args.data
+    if (typeof object === 'string') {
```
nklomp, line 105: This should probably not be here

nklomp (changes requested):
See remarks. Please also provide some examples in the README. I didn;'t get around to those for the SIOP QRs as I simply didn't have the time to get everything in there in time for the demo

packages/qr-code-generator/src/types/IQRCodeGenerator.ts
maikel-maas: Fixed!

packages/qr-code-generator/src/types/IQRCodeGenerator.ts
maikel-maas: Fixed!

packages/qr-code-generator/src/agent/QrCodeProvider.tsx
```diff
@@ -87,33 +87,29 @@ export class QrCodeProvider implements IAgentPlugin {
...
     }
     const scheme = (object.scheme ?? 'openid-credential-offer://').replace('://?', '').replace('://', '') + '://'
-    const domain = object.domain ?? ''
-    const path = object.path ?? ''
+    const baseUri = object.baseUri ?? ''
     const credentialOfferUri = `?credential_offer_uri=${object.credentialOfferUri?.replace('credential_offer_uri=', '')}`
     const credentialOffer = `?credential_offer=${object.credentialOffer?.replace('credential_offer=', '')}`
     const value = object.credentialOffer ? credentialOffer : credentialOfferUri
-    return `${scheme}${domain}${path}${value}`
+    return `${scheme}${baseUri}${value}`
```
nklomp, line 113: hmzz, I think we need to have some logic which tests if a scheme is present or not. a BaseURI should already start with the scheme normally.

It makes sense to ensure that we are not getting 2 schemes in the eventual URI when someone both uses the scheme and also includes the same scheme in the baseUri as they are used to a URI having a scheme

Simply doublechecking whether the scheme value (if present) is not already part of the baseUri before the concatenation would suffice for now

nklomp (changes requested):
Small remark about scheme and baseUri

We need to doublecheck whether the scheme value (if present) is not already part of the baseUri before the concatenation would suffice for now

packages/qr-code-generator/src/agent/QrCodeProvider.tsx
maikel-maas: Fixed!

packages/qr-code-generator/src/agent/QrCodeProvider.tsx
```diff
@@ -107,6 +107,9 @@ export class QrCodeProvider implements IAgentPlugin {
     }
     const scheme = (object.scheme ?? 'openid-credential-offer://').replace('://?', '').replace('://', '') + '://'
     const baseUri = object.baseUri ?? ''
+    if (object.scheme && baseUri.includes(object.scheme)) {
```
nklomp, line 110: This is not doing what I suggested

I am okay if they are the same.
I a not okay if they differ.

So a check on whether the scheme is part of it. If so ignore it. If there is a scheme and the base URI does not contain a scheme (check for ://) it should be appended. If both are present and they differ an error should be thrown

nklomp (changes requested):
See remarks.

packages/qr-code-generator/src/agent/QrCodeProvider.tsx
maikel-maas: Fixed!

maikel-maas: I think it's fixed now

PR #96, opened 2023-03-21
VDX-175: velocity-message-handler moved to ssi-plugins

sanderPostma:


sanderPostma:
> This is not what should happen. Something else is going on if this happens

Reverted that. Looks like it's been going on for a while in other branches as well. I only deleted the Velocity message handler, no other changes now.

nklomp (approved):

PR #98, opened 2023-03-27
Feature/vdx 173 schema gen

sanderPostma:


packages/vc-handler-ld-local/src/ld-suites.ts
```diff
@@ -1,5 +1,5 @@
 import { CredentialPayload, IAgentContext, IKey, IKeyManager, IResolver, PresentationPayload, TKeyType, VerifiableCredential } from '@veramo/core'
-import { DIDDocument } from 'did-resolver/src/resolver'
+import { DIDDocument } from 'did-resolver/lib/resolver'
```
sanderPostma, line 2: the schema generator can't use imports from src, it needs *.d.ts files

nklomp (approved):
LGTM

PR #104, opened 2023-04-14
Siop verifier cleanup

sksadjad:
deleted/changed console logs with debugs

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,6 @@
     "build:clean": "lerna clean -y && lerna bootstrap && lerna run build:clean --concurrency 1 && yarn build:copyfiles && yarn build:schema-api",
     "build:schema-api": "lerna run extract-api && lerna run generate-plugin-schema",
     "build:copyfiles": "lerna run build:copyfiles",
-    "clean": "rimraf --glob **/dist **/coverage packages/**/node_modules node_modules packages/**/tsconfig.tsbuildinfo",
     "bootstrap": "lerna bootstrap",
     "test:ci": "jest --config=jest.json",
     "test": "jest --verbose --config=jest.json --coverage=true --detectOpenHandles",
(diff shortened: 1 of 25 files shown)


package.json
```diff
@@ -11,6 +11,7 @@
     "build:clean": "lerna clean -y && lerna bootstrap && lerna run build:clean --concurrency 1 && yarn build:copyfiles && yarn build:schema-api",
     "build:schema-api": "lerna run extract-api && lerna run generate-plugin-schema",
     "build:copyfiles": "lerna run build:copyfiles",
+    "clean": "rimraf --glob **/dist **/coverage packages/**/node_modules node_modules packages/**/tsconfig.tsbuildinfo",
```
nklomp, line 14: There already is a build-clean doing roughly the same

packages/siopv2-oid4vp-rp-rest-api/src/SIOPv2RPRestAPI.ts
```diff
@@ -241,14 +239,16 @@ export class SIOPv2RPRestAPI {
...
           .then((verifiedResponse: VerifiedAuthorizationResponse) => {
-            // console.log('verifiedResponse: ', JSON.stringify(verifiedResponse, null, 2))
-
             const wrappedPresentation = verifiedResponse?.oid4vpSubmission?.presentations[0]
             if (wrappedPresentation) {
               const credentialSubject = wrappedPresentation.presentation.verifiableCredential[0].credential.credentialSubject
-              console.log('AND WE ARE DONE!')
-              console.log(JSON.stringify(credentialSubject, null, 2))
-              console.log(JSON.stringify(wrappedPresentation.presentation, null, 2))
+              debug(
```
nklomp, line 245: Please revert this one. This module is still being developed and tested. I will do a pass once it is ready

nklomp (changes requested):
Requesting changes to block the merge. As the module still is seeing heavy development the most console.log lines are making sense.

In the mean time there are a few smaller comments like for instance the clean option you added, which can be addressed

package.json
sksadjad: removed

sksadjad:
I've added args for SIOPv2OID4VPRestClient constructor in this pr

PR #110, opened 2023-06-05
feature/WAL-617

BtencateSphereon:


packages/data-store/src/entities/issuanceBranding/ImageAttributesEntity.ts
```diff
@@ -0,0 +1,57 @@
...
+  @Column({ name: 'uri', length: 255, nullable: true, unique: false })
+  @Validate(IsNonEmptyStringConstraint, { message: 'Blank image uri are not allowed' })
+  uri?: string
+
+  @Column({ name: 'base64Content ', length: 255, nullable: true, unique: false })
+  @Validate(IsNonEmptyStringConstraint, { message: 'Blank image base64 content are not allowed' })
+  base64Content?: string
+
+  @Column({ name: 'type', length: 255, nullable: true, unique: false })
+  @Validate(IsNonEmptyStringConstraint, { message: 'Blank image types are not allowed' })
```
nklomp, line 21: Maybe change the name 'type' to mediaType, as that is what it is

packages/data-store/src/entities/issuanceBranding/ImageDimensionsEntity.ts
```diff
@@ -0,0 +1,22 @@
+import { BaseEntity, Column, Entity, PrimaryGeneratedColumn } from 'typeorm'
+import { IBasicImageDimensions } from '../../types'
+
+@Entity('ImageDimensions')
+export class ImageDimensionsEntity extends BaseEntity {
+  @PrimaryGeneratedColumn('uuid')
+  id!: string
+
+  @Column({ name: 'width', nullable: false, unique: false })
+  width!: number
```
nklomp, line 10: See other remark. Probably it makes sense to change these to string, if we want to support other dimension types in the future (proportional, percentage etc)

packages/data-store/src/entities/validators/index.ts
```diff
@@ -0,0 +1,16 @@
+import { ValidationArguments, ValidatorConstraint, ValidatorConstraintInterface } from 'class-validator'
+
+@ValidatorConstraint({ name: 'isNonEmptyString', async: false })
+export class IsNonEmptyStringConstraint implements ValidatorConstraintInterface {
```
nklomp, line 4: This one is being used quite heavily. Given that a lot of the values actually are populated from external metadata sources, this could result in us throwing errors, simply because for instance an issuer had an empty string for an attribute. If we want to keep this everywhere, it means we will have to check in the mapping process from external data to the entity, whether a value is empty. If it is then store it as a null value. We should at least document that then

packages/data-store/src/entities/issuanceBranding/ImageDimensionsEntity.ts
nklomp: Okay, guess this always comes from the absolute dimensions of an image, so scratch that

nklomp (commented):
See remarks. Lot of code, but on first quick glance looks okay. See the point about non-empty string requirements in most entity properties

packages/data-store/src/entities/validators/index.ts
BtencateSphereon: Yes, that is a fair point. The reason I added this validation is to not have to deal with those values on the end, and a user can just check if the value is present. 

I have changed the code to make them undefined when saving them to the db

nklomp:
PR approved

nklomp (approved):
LGTM

PR #115, opened 2023-07-04
Feature/vdx 238 fix validated creds

sanderPostma:
Please check, you mean like this? It is working but now verifiedData is completely hidden from the types.


nklomp:
I am okay with it being available as an optional property in the SDK. So you could simply extend an existing interface from the SIOP lib for instance. I am not okay with having that type in the lower-level SIOP lib, since these types/payloads need to be spec compliant.

I am not a big fan of just adding some additional claims in some areas, not really visible to the outside world in the SDK, because then values come in as a surprise in some situations.

sanderPostma:
@nklomp more like this?
 Added verifiedData to AuthStatusResponse (authmodel) and created AuthorizationResponseStateWithVerifiedData in ISIOPv2RP
 
 (Do not merge yet, I still have to test it after the modifications)

nklomp (approved):

PR #123, opened 2023-08-21
feature/VDX-247 added a module for contact manager

sksadjad:


nklomp (changes requested):
Please look at the requirements. None of them are being met....

packages/contact-manager-rest-api/__tests__/agent.ts
```diff
@@ -0,0 +1,17 @@
+import { createAgent } from '@veramo/core'
+import { DataStore, DataStoreORM } from '@veramo/data-store'
+import { IRequiredPlugins } from '../src'
+import { DB_CONNECTION_NAME, getDbConnection } from './database'
+import { ContactManager } from '@sphereon/ssi-sdk.contact-manager'
+import { ContactStore } from '@sphereon/ssi-sdk.data-store'
+
+
+export const baseUrl = 'https://sphereon.cm.ngrok.dev'
```
nklomp, line 9: Why is ngrok being used?

packages/contact-manager-rest-api/__tests__/database/databaseService.ts
```diff
@@ -0,0 +1,36 @@
+import Debug from 'debug'
```
nklomp, line 1: ssi-agent has a class DataSources that handles all of this these days.

packages/contact-manager-rest-api/package.json
```diff
@@ -0,0 +1,81 @@
...
+    "@sphereon/ssi-sdk.contact-manager": "workspace:^",
+    "@sphereon/ssi-sdk.data-store": "workspace:^",
+    "@sphereon/ssi-sdk-ext.key-manager": "0.14.0",
+    "@sphereon/ssi-sdk-ext.key-utils": "0.14.0",
+    "@sphereon/ssi-sdk.core": "workspace:^",
+    "@sphereon/ssi-sdk.kv-store-temp": "workspace:^",
+    "@sphereon/ssi-types": "workspace:^",
+    "@veramo/core": "4.2.0",
+    "body-parser": "^1.19.0",
+    "casbin": "^5.26.2",
```
nklomp, line 24: casbin and passport-http-bearer and morgan should move to dev deps

packages/contact-manager-rest-api/src/api-functions.ts
```diff
@@ -0,0 +1,127 @@
+import { checkAuth, sendErrorResponse, ISingleEndpointOpts } from '@sphereon/ssi-express-support'
+import { Request, Response, Router } from 'express'
+import { IRequiredContext } from './types'
+import Debug from 'debug'
+import { AddContactArgs } from '@sphereon/ssi-sdk.contact-manager'
+
+const debug = Debug('sphereon:ssi-sdk:contact-manager-rest-api')
+
+export function partyReadEndpoints(router: Router, context: IRequiredContext, opts?: ISingleEndpointOpts) {
```
nklomp, line 9: Please keep the exact structure used elsewhere. There is a reason why there is a function per endpoint! Namely to allow others that do not want to use the class enabling all functions/endpoints. This allows any party to just integrate one or more endpoints in their solution if they desire. With the approach below they cannot decide anymore which endpoints to include, because you will get multiple. 

We will combine certain endpoints in the features that the class is enabling/disabling

packages/contact-manager-rest-api/src/api-functions.ts
```diff
@@ -0,0 +1,127 @@
...
+  if (opts?.enabled === false) {
+    debug(`Endpoint is disabled`)
```
nklomp, line 11: This says nothing and the console.logs were deliberate

packages/contact-manager-rest-api/src/api-functions.ts
```diff
@@ -0,0 +1,127 @@
...
+  }
+  const path = opts?.path ?? '/parties'
+  router.post(path, async (request: Request, response: Response) => {
+    try {
+      const addParty = request.body
+      const party = await context.agent.cmAddContact(addParty as AddContactArgs)
+      response.statusCode = 201
+      return response.send(party)
+    } catch (error) {
+      console.error(error)
```
nklomp, line 55: No console.errors. The sendErrorResponse already takes care of that

packages/contact-manager-rest-api/src/api-functions.ts
```diff
@@ -0,0 +1,127 @@
...
+      return sendErrorResponse(response, 500, 'Could not add party')
```
nklomp, line 56: Please pass the exceptions to sendErrorResponse

packages/contact-manager-rest-api/src/contact-manager-api-server.ts
```diff
@@ -0,0 +1,85 @@
...
+  private readonly _agent: TAgent<IRequiredPlugins>
+  private readonly _opts?: IContactManagerAPIEndpointOpts
+  private readonly _router: Router
+
+  constructor(args: { agent: TAgent<IRequiredPlugins>; expressSupport: ExpressSupport; opts?: IContactManagerAPIEndpointOpts }) {
+    const { agent, opts } = args
+    this._agent = agent
+    copyGlobalAuthToEndpoints({ opts, keys: ['contactRead', 'contactTypeRead', 'identityRead'] })
+    const enableFeatures = opts?.enableFeatures ?? [
+      ContactManagerMRestApiFeatureEnum.party_read,
```
nklomp, line 25: Do not start incorporating enums all of a sudden. The strings where typed to begin with, so you cannot just have a string that is not in the type.
Enums also do not play really nice with payloads and/or env vars. Please ask questions if you see me doing one thing and you start doing another thing. There is always are reason why I do something!

packages/contact-manager-rest-api/src/contact-manager-api-server.ts
```diff
@@ -0,0 +1,85 @@
...
+      ContactManagerMRestApiFeatureEnum.party_write,
+      ContactManagerMRestApiFeatureEnum.party_type_read,
+      ContactManagerMRestApiFeatureEnum.party_type_write,
+      ContactManagerMRestApiFeatureEnum.identity_read,
+      ContactManagerMRestApiFeatureEnum.identity_write,
+    ]
+    console.log(`Contact Manager API enabled, with features: ${JSON.stringify(features)}}`)
+
+    // todo: I've commented out the write part of the APIs. We might want to implement and uncomment these in near future
```
nklomp, line 46: Remarks is not in line with code

packages/contact-manager-rest-api/src/request-agent-router.ts
```diff
@@ -0,0 +1,47 @@
+import { IAgent } from '@veramo/core'
```
nklomp, line 1: Class/file is not needed

packages/contact-manager-rest-api/src/contact-manager-api-server.ts
```diff
@@ -0,0 +1,85 @@
...
+export class ContactManagerApiServer {
+  private readonly _express: Express
+  private readonly _agent: TAgent<IRequiredPlugins>
+  private readonly _opts?: IContactManagerAPIEndpointOpts
+  private readonly _router: Router
+
+  constructor(args: { agent: TAgent<IRequiredPlugins>; expressSupport: ExpressSupport; opts?: IContactManagerAPIEndpointOpts }) {
+    const { agent, opts } = args
+    this._agent = agent
+    copyGlobalAuthToEndpoints({ opts, keys: ['contactRead', 'contactTypeRead', 'identityRead'] })
```
nklomp, line 23: This should include all potential configs used by the endpoints. Since you enabled write endpoints, it should also include these configs

nklomp (changes requested):
See remarks. Some are applicable to multiple parts of the code.

packages/contact-manager-rest-api/__tests__/agent.ts
sksadjad: deleted the baseurl

packages/contact-manager-rest-api/package.json
sksadjad: moved

packages/contact-manager-rest-api/src/api-functions.ts
sksadjad: changed to method per endpoint

packages/contact-manager-rest-api/src/api-functions.ts
sksadjad: changed to console.log

packages/contact-manager-rest-api/src/api-functions.ts
sksadjad: removed

packages/contact-manager-rest-api/src/api-functions.ts
sksadjad: passed

packages/contact-manager-rest-api/src/contact-manager-api-server.ts
sksadjad: removed the enum and used literal strings

packages/contact-manager-rest-api/src/contact-manager-api-server.ts
sksadjad: removed

packages/contact-manager-rest-api/src/request-agent-router.ts
sksadjad: deleted

packages/contact-manager-rest-api/src/contact-manager-api-server.ts
sksadjad: changed to include all

packages/contact-manager-rest-api/__tests__/database/databaseService.ts
sksadjad: changed debug to console logs

packages/contact-manager-rest-api/src/contact-manager-api-server.ts
```diff
@@ -0,0 +1,85 @@
...
+      'party_read',
+      'party_write',
+      'party_type_read',
+      'identity_read',
+    ]
+    this._opts = opts
+    this._express = args.expressSupport.express
+    this._router = express.Router()
+    const context = agentContext(agent)
+    const features = enableFeatures ?? [
```
nklomp, line 41: What's up with this. Above you are already doing enabledFeatures with a nullish coalescence. This one will never be triggered since it cannot be null at this point. Also why would the defaults differ between the two?

packages/contact-manager-rest-api/src/types.ts
```diff
@@ -0,0 +1,24 @@
+import { GenericAuthArgs, ISingleEndpointOpts } from '@sphereon/ssi-express-support'
+import { IContactManager } from '@sphereon/ssi-sdk.contact-manager'
+import { IAgentContext, IDataStore, IKeyManager } from '@veramo/core'
+
+export type ContactManagerMRestApiFeatures =
+  'party_read' |
+  'party_write' |
+  'party_type_read' |
+  'identity_read'
```
nklomp, line 9: This has less options than what is being set in the RestAPI on the features array (which will never be set, given current logic, but still). Is this type missing features, or is the RestAPI trying to set features that do no exist (yet)

packages/contact-manager-rest-api/__tests__/database/test.sqlite
nklomp, line 1: Remove from git

nklomp (changes requested):
See remarks

packages/contact-manager-rest-api/__tests__/database/test.sqlite
sksadjad: removed

packages/contact-manager-rest-api/src/types.ts
sksadjad: fixed the logic and changed the `contact-manager-api-server`

packages/contact-manager-rest-api/__tests__/database/config.ts
```diff
@@ -0,0 +1,20 @@
...
+import { SqliteConnectionOptions } from 'typeorm/driver/sqlite/SqliteConnectionOptions'
+
+const DB_CONNECTION_NAME = 'default'
+const DB_ENCRYPTION_KEY = '29739248cad1bd1a0fc4d9b75cd4d2990de535baf5caadfdf8d8f86664aa830c'
+
+const sqliteConfig: SqliteConnectionOptions = {
+  type: 'sqlite',
+  database: '__tests__/database/test.sqlite',
+  entities: [...VeramoDataStoreEntities, ...DataStoreContactEntities],
+  migrations: [...VeramoDataStoreMigrations, ...DataStoreMigrations],
```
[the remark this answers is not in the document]
sksadjad: changed this to:
`entities: [...DataStoreContactEntities],
  migrations: [...DataStoreMigrations],`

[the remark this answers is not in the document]
nklomp: No, the issue is not with the Varamo bits; they should be there. Look at the name of the entities vs the migrations. You are only including the ContactEntities (these are not all the entities the datastore exports, but then are doing all the migrations the datastore exposes)

packages/contact-manager-rest-api/__tests__/database/databaseService.ts
sksadjad: changed this to use DataSources from `@sphereon/ssi-sdk.agent-config` instead and therefore removed the extra class `databaseService`

packages/contact-manager-rest-api/__tests__/RestAPI.ts
```diff
@@ -1,43 +1,48 @@
...
+new ContactManagerApiServer({
+  opts: {
+    endpointOpts: {
+      globalAuth: {
+        authentication: {
+          enabled: false,
+        },
+      },
+      partyRead: {
+        disableGlobalAuth: true
```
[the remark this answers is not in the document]
nklomp: If you don't want authentication, then do not set it up in the first place with passport. Leave all of that out then this is also no need to set all these options etc.

nklomp:
LGTM

nklomp (approved):

PR #126, opened 2023-09-13
feat: added contact test data

sksadjad:


nklomp (approved):

nklomp (changes requested):
Please make these values more in line with what was used in the Web-walet so we have a bit of a feal what more realistic values look like in the ui

sksadjad:
@nklomp changed the test values to be inline with previous version of test data

nklomp (approved):

PR #128, opened 2023-09-19
Added identies to contacts

sanderPostma:


packages/contact-manager-rest-api/__tests__/agent.ts
```diff
@@ -14,48 +21,70 @@ const dbConnection = DataSources.singleInstance()
 const agent = createAgent<IRequiredPlugins>({
   plugins: [new DataStore(dbConnection), new DataStoreORM(dbConnection), new ContactManager({ store: new ContactStore(dbConnection) })],
 })
+const generateIdentity = (contact: Record<string, any>) : NonPersistedIdentity => {
```
nklomp, line 24: Please use the agent method to generate a key and never use PII in DIDs (even though this is not stored on a ledger, never use PII for them)

nklomp (changes requested):
Please use the agent method to generate a didKey for you, then assign the did value to the contact. So we can actually use these DIDs from the agent as well. Also never do PII in DIDs!

didKeyIdentifier = await agent.didManagerCreate() // you might need to set the method if did:key isn't defined as the default method on the agent.

sanderPostma:
creating did:ion now
Also includes a fix that Bram made remotely after I ran into a typeorm issue.

nklomp (approved):

PR #144, opened 2023-11-25
Feature/dpp 90

sksadjad:
added the ssi-sdk.core function for `enablePostgresUuidExtension` and exported it. also made data-store use it


packages/data-store/package.json
```diff
@@ -13,6 +13,7 @@
     "typeorm-postgres:migration:run": "pnpm run typeorm -- migration:run -c migration-postgres"
   },
   "dependencies": {
+    "@sphereon/ssi-sdk.core": "workspace:*",
```
nklomp, line 16: Pleases also update the references in tsconfig.json

nklomp (changes requested):
Please update tsconfig

packages/ssi-sdk-core/src/utils/database.ts
```diff
@@ -1,3 +1,20 @@
 export const flattenArray = <T>(args: { items: Array<T | Array<T>> }): Array<T> => args.items.flat() as Array<T>
 
 export const flattenMigrations = <T>(args: { migrations: Array<T | Array<T>> }): Array<T> => args.migrations.flat() as Array<T>
+
+/**
+ * It should accept queryRunner from the typeorm
+ */
+export const enablePostgresUuidExtension = async (queryRunner: any) => {
+  if (!queryRunner.query) {
```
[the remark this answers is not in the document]
sksadjad: modified this to have a type which has query function

packages/data-store/package.json
sksadjad: added core to tsconfig of data-store

packages/ssi-sdk-core/src/utils/database.ts
```diff
@@ -2,19 +2,18 @@ export const flattenArray = <T>(args: { items: Array<T | Array<T>> }): Array<T>
 
 export const flattenMigrations = <T>(args: { migrations: Array<T | Array<T>> }): Array<T> => args.migrations.flat() as Array<T>
 
-/**
- * It should accept queryRunner from the typeorm
```
nklomp: Okay an now add the remark back to the type please, so people know you are creating a type that mimics TypeOrm

nklomp (approved):

PR #145, opened 2023-11-28
feat(ssi-types): sd-jwt support

TimoGlastra:
~~Still draft as I need to fix some tests, and the packages in this repo dependant on SSI types, but would like to submit for initial review already to get some feedback on the approach.~~

Adding SD-JWT VC support to all interface feels a bit 'hacky' as it's completely different from the W3C data model. I've thought about creating another abstraction layer, so there's better distinction and separation between W3C and SD-JWT VCs, but as the spec is still really new, and I don't have a lot of experience with it yet, it felt wrong to do that at this point.

~~There's still some improvements to be made to the VC / VP handling, as the SD-JWT is both the VC as well as the VP (with the addition of the KB-JWT).~~

~~And I want to add support for decoding of an SD-JWT string into the decoded interface format.~~


nklomp:
Yeah agree. I think it makes sense we thing about when the interfaces for the MDL/MDOC land as well, as there is some degree of overlap there as well.

nklomp:
Overall looking good. It might be worthwhile to place some comments in the code where the original values being returned. As for the other formats these are the actual payloads received initially like the JWT string or the JSON-LD VC. But that is basically what your last sentence above is also about

TimoGlastra:
@nklomp fixed the tests, added support for decoding of SD-JWT credentials, and also added a method to decode an SD-JWT async.

There's quite some sync hashers available, and if you use those you can pass the sync hasher directly to all the existing methods of the credential mapper. If it's async you must call the async decode method.

It's currently based on the `@sd-jwt/core` package (@berendsliedrecht has already moved it to a monorepo), and our plan is to create a separate `@sd-jwt/decode` package that we can use here (like `jwt_decode`). 

However, if for now we can already merge it based on the full library and we swap it in a future PR that will help us a lot in making some progress with the different packages. The methods are created in such a way the the public interface does not depend at all on the `@sd-jwt/core` library, and so we can easily swap it for `@sd-jwt/decode` or another library without it being a breaking change.

nklomp:
Sure that is fine (full instead of separate packages for now). Will have a look at the PR later today

TimoGlastra:
Hey @nklomp have you had a chance to look at this PR? We have some dependencies on these libraries (ssi-types, oid4vci, pex, oid4vp-siop) and need them to be merged/released before we can get this finalized in AFJ as well. 

We can also look at creating forked released of the libraries in the meantime on our side, so you have all the time to review and merge the SD-JWT support, but this would of course introduce overhead that we'd rather avoid if possible

nklomp:
Sorry have been a bit under the weather last week and thought this was already merged. Will have a look at it now

nklomp (approved):

TimoGlastra:
Thanks @nklomp, hope you're feeling better!

PR #150, opened 2024-01-11
feat(ssi-types): add kid to cnf

TimoGlastra:
Small update to support `kid` in an sd-jwt `cnf` claim, allowing the sd-jwt to be bound to a did in addition to the already supported `jwk`


nklomp (approved):
LGTM

PR #152, opened 2024-01-17
chore: update sd-jwt

TimoGlastra:
Updates the SD-JWT library. Draft for now as we first want to create a stable release in SD-jWT so we can depend on it here

~~We split up the SD-jWT library now so that (after https://github.com/berendsliedrecht/sd-jwt-ts/pull/29 is merged) we only have to depend on `@sd-jwt/decode` to decode the sd-jwt credential~~

~~For PEX we still have to depend on `@sd-jwt/core` for now, but in the future we can probably split up the parts needed in pex into a separate `@sd-jwt/present` package (cc @berendsliedrecht)~~

### DIFF
--- a/packages/ssi-types/__tests__/encoding.test.ts
+++ b/packages/ssi-types/__tests__/encoding.test.ts
@@ -1,6 +1,9 @@
+import { createHash } from 'crypto'
 import * as fs from 'fs'
 import {
   CredentialMapper,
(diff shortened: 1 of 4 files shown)


nklomp (approved):

TimoGlastra:
Lol ninja 🥷 @nklomp

PR #156, opened 2024-01-22
Including dynamic token for endpoint protection

### LINKED ISSUE #142: Including dynamic token for endpoint protection
The authentication option with a static token is good for demo purposes or for initialisation a veramo agent on demand for each call. But in my case I am using auth0 to add an access token that is valid for some minutes, longer than the user is interacting with the agent.

So extending the type like:
```
export interface IRestClientAuthenticationOpts {
    enabled?: boolean;
    staticBearerToken?: string;
    dynamicBearerToken?: () => Promise<string>;
}
```
And then update the OID4VCIRestClient like:
```
  private async createHeaders(existing?: Record<string, any>): Promise<HeadersInit> {
    const headers: HeadersInit = {
      ...existing,
      Accept: 'application/json',
    }
    if (this.authOpts?.enabled === true) {
      if (!this.authOpts.staticBearerToken || !this.authOpts.dynamicBearerToken) {
        throw Error(`Cannot have authentication enabled, whilst not enabling static bearer or a dynamic tokens at this point`)
      }
      if(this.authOpts.staticBearerToken && this.authOpts.dynamicBearerToken !== undefined) {
        throw Error(`Please either pass a static token or a dynamic token, not both.`);
      }
      if(this.authOpts.dynamicBearerToken !== undefined) {
        headers.Authorization = `Bearer ${await this.authOpts.dynamicBearerToken}`
      } else {
        headers.Authorization = `Bearer ${this.authOpts.staticBearerToken}`
      }
    }
    return headers
  }
  ```

To get more flexibility I think making `createHeaders` has to be async and the two calls in the class have to be updated

Should solve the problem. What do you think about the proposal?

cre8:
Fixes #142

Adding the dynamic token also to the SIOPv2OID4VPRPRestClient

### DIFF
--- a/packages/siopv2-oid4vp-rp-rest-client/src/agent/SIOPv2OID4VPRPRestClient.ts
+++ b/packages/siopv2-oid4vp-rp-rest-client/src/agent/SIOPv2OID4VPRPRestClient.ts
@@ -33,7 +33,7 @@ export class SIOPv2OID4VPRPRestClient implements IAgentPlugin {
     this.authOpts = args?.authentication
   }
 
-  private createHeaders(existing?: Record<string, any>): HeadersInit {
+  private async createHeaders(existing?: Record<string, any>): Promise<HeadersInit> {
     const headers: HeadersInit = {
       ...existing,
       Accept: 'application/json',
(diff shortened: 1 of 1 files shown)


nklomp (approved):
LGTM, thx

PR #157, opened 2024-02-02
feature/VDX-318

BtencateSphereon:


packages/event-logger/src/agent/EventLogger.ts
```diff
@@ -29,7 +33,7 @@ export class EventLogger implements IAgentPlugin {
   public async onEvent(event: LoggingEvent, context: RequiredContext): Promise<void> {
     switch (event.type) {
       case LoggingEventType.AUDIT:
-        await this.loggerLogAuditEvent({ event: event.data }, context)
+        await context.agent.loggerLogAuditEvent({ event: event.data })
```
nklomp, line 36: Probably wise to add a comment why it is calling the agent instead of this (to make sure it uses the rest client if configured)

nklomp (commented):
Small remark about a missing comment, which is vital to understand for other developers

packages/event-logger/src/agent/EventLogger.ts
BtencateSphereon: added a comment

sksadjad (approved):
LGTM

PR #159, opened 2024-02-09
Fix/dpp 125

sksadjad:
fixed the problem with image-size. the version of image-size previously used was not react friendly


packages/ssi-sdk-core/src/utils/image.ts
```diff
@@ -1,25 +1,67 @@
...
 import { IImageDimensions, IImageResource } from '../types'
 
+type SizeCalculationResult = {
+  width?: number
+  height?: number
+  orientation?: number
+  type?: string
+}
+
+const uint8ArrayToString = (uint8Array: Uint8Array): string => {
```
nklomp, line 12: Please use the uint8arrays package for that as we are using that in everywhere. It has a toString method with encoding support

packages/ssi-sdk-core/src/utils/image.ts
```diff
@@ -1,25 +1,67 @@
...
+  const decoder = new TextDecoder('utf-8')
+  return decoder.decode(uint8Array)
+}
+
+// TODO: here we're handling svg separately, remove this section when image-size starts supporting it in version 2
+const isSvg = (uint8Array: Uint8Array): boolean => {
+  const text = uint8ArrayToString(uint8Array)
```
nklomp, line 19: Can't we detect this based in the array itself, or at least only take a few indices first. This probably has a massive hit on performance. This for instance will covert 10 MB images into text first, only to determine first whether it is a SVG or not

packages/ssi-sdk-core/src/utils/image.ts
```diff
@@ -1,25 +1,67 @@
...
+  const normalizedText = text.trim().toLowerCase()
+  return normalizedText.startsWith('<svg') || normalizedText.startsWith('<?xml')
+}
+
+const getSvgDimensions = (uint8Array: Uint8Array) => {
+  const svgContent = uint8ArrayToString(uint8Array)
+  const widthMatch = svgContent.match(/width="([^"]+)"/)
```
nklomp, line 26: Does this work at all? For a simple SVG probably it might, but what about an SVG with all kinds of objects in it?

packages/ssi-sdk-core/src/utils/image.ts
```diff
@@ -1,25 +1,67 @@
...
+  const heightMatch = svgContent.match(/height="([^"]+)"/)
+  const viewBoxMatch = svgContent.match(/viewBox="([^"]+)"/)
+
+  let width, height
```
nklomp, line 30: Why are these not typed?

packages/ssi-sdk-core/src/utils/image.ts
```diff
@@ -1,25 +1,67 @@
...
+  }
+
+  if (heightMatch) {
+    height = heightMatch[1]
+  }
+
+  if ((!width || !height) && viewBoxMatch) {
+    const parts = viewBoxMatch[1].split(' ').map(Number)
+    if (!width && parts.length === 4) {
+      width = parts[2].toString()
```
nklomp, line 43: Why a tostring on something that is mapped to a number?

packages/ssi-sdk-core/src/utils/image.ts
```diff
@@ -1,25 +1,67 @@
...
+    return `image/svg+xml`
   }
+  const result: SizeCalculationResult = imageSize(buffer)
+  return `image/${result.type}`
 }
 
 export const getImageDimensions = async (base64: string): Promise<IImageDimensions> => {
   const buffer: Buffer = Buffer.from(base64, 'base64')
-  const dimensions: ISizeCalculationResult = sizeOf(buffer)
+  const dimensions: SizeCalculationResult = isSvg(buffer) ? getSvgDimensions(buffer) : imageSize(buffer)
```
nklomp, line 64: Just create a getImageDimensions function, so other places can also reuse it

nklomp (changes requested):
See remarks. Espacially the toString is too of a naive implementation when working with images

packages/ssi-sdk-core/src/utils/image.ts
sksadjad: changed to use that library

packages/ssi-sdk-core/src/utils/image.ts
sksadjad: yeah, changed the code to use only first bytes

packages/ssi-sdk-core/src/utils/image.ts
sksadjad: changed this method and wrote 2 additional tests for a multi-layered and a animated svg

packages/ssi-sdk-core/src/utils/image.ts
sksadjad: changed these and made everything typed

packages/ssi-sdk-core/src/utils/image.ts
sksadjad: changed this, width/height is a number now

packages/ssi-sdk-core/src/utils/image.ts
sksadjad: we already have `getImageDimensions` if the image is svg, we're calling a special method called `getSvgDimensions` which is not exported. Do you want this to be also exported?

nklomp: Okay then it does make sense to also add an uint8array next to the string for this function

packages/ssi-sdk-core/src/utils/image.ts
```diff
@@ -1,25 +1,61 @@
...
-      return `image/${result.type}`
+  if (isSvg(buffer)) {
+    return `image/svg+xml`
   }
+  const result: SizeCalculationResult = imageSize(buffer)
+  return `image/${result.type}`
 }
 
 export const getImageDimensions = async (base64: string): Promise<IImageDimensions> => {
   const buffer: Buffer = Buffer.from(base64, 'base64')
```
nklomp, line 57: We avoid Buffer by default because of RN. Use U8Int Array which is compatible

packages/ssi-sdk-core/src/utils/image.ts
sksadjad: now it's accepting both base64 string and uint8array. also added explanation at the top of the method

sksadjad: also applied this to the `getImageMediaType` function. now it's also accepting `uint8array` type

nklomp (approved):

PR #160, opened 2024-02-16
Feature/sdk 3

maikel-maas:


packages/xstate-persistence/src/types/types.ts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+export enum SQLDialect {
+  SQLite3 = 'SQLite3',
+  PostgreSQL = 'PostgreSQL',
+}
+
+export type DeleteExpiredStatesArgs = {
+  duration: number
+  dialect: SQLDialect
```
nklomp, line 19: This makes no sense to me. A caller should not have to know any implementation detail like this

packages/data-store/src/types/xstate/IAbstractXStateStore.ts
```diff
@@ -0,0 +1,47 @@
+export type SaveStateArgs = {
+  step: string
+  type: string
+  eventName: string
+  state: string
```
nklomp, line 5: I am not to happy about the caller having to worry about transforming the state to a string first. Ideally that happens in the plugin itself near the persistence layer. It would be nice if we can figure out whether the state itself from Xstate is already serializable (probably it is). If that is the case then I rather have a mapper internally close to the persistence or datastore layer that does the (de)serialization, so the outside world can just pas in Xstate states. Of course we would have to look at some generic arguments, because we don't want to pull in Xstate at the datastore plugin, so either we would have to recreate some types/interfaces, or pull it in at the Xstate Plugin level. If states however are not really serializable by default, then ideally we do provide some mapper functions in the Xstate Plugin, so that a caller doesn't need to worry about how to (de)serialize to much. Even it is a simple JSON.parse/stringify, I rather see some methods calling JSON.parse/stringify

packages/data-store/src/xstate/XStateStore.ts
```diff
@@ -0,0 +1,86 @@
...
+      debug(`Error deleting state: ${error}`)
+      return false
+    }
+  }
+
+  async deleteExpiredStates(args: DeleteExpiredStateArgs): Promise<boolean> {
+    try {
+      const connection: DataSource = await this.dbConnection
+      debug(`Executing deleteExpiredStates query with params: ${JSON.stringify(args)}`)
+      let deleteCriteria: FindOptionsWhere<StateEntity> = { expiresAt: LessThan(new Date()) }
```
nklomp, line 66: This can be done nicer with a const and then ...(args.type && {type: args.type})

packages/data-store/src/entities/xstate/StateEntity.ts
```diff
@@ -0,0 +1,47 @@
...
+  @PrimaryGeneratedColumn('uuid', { name: 'id' })
+  id!: string
+
+  @Column({ name: 'step', nullable: false })
+  step!: string
+
+  @Column({ name: 'type', nullable: false })
+  type!: string
+
+  @Column({ name: 'event_name', nullable: false })
```
nklomp, line 15: Elsewhere it is called an XStateEventType, which makes more sense to me

packages/data-store/src/entities/xstate/StateEntity.ts
```diff
@@ -0,0 +1,47 @@
+import { BaseEntity, Column, CreateDateColumn, Entity, PrimaryGeneratedColumn, UpdateDateColumn } from 'typeorm'
+import { NonPersistedXStateStoreEvent } from '../../types'
+
+@Entity('StateEntity')
+export class StateEntity extends BaseEntity {
+  @PrimaryGeneratedColumn('uuid', { name: 'id' })
+  id!: string
```
nklomp, line 7: This should not be generated, as every Xstate instance already has a unique id AFAIK

packages/data-store/src/entities/xstate/StateEntity.ts
```diff
@@ -0,0 +1,47 @@
...
+  @Column({ name: 'completed_at', type: 'datetime', nullable: true })
+  completedAt?: Date
+
+  @Column({ name: 'tenant_id', type: 'varchar', nullable: true })
+  tenantId?: string
+}
+
+export const stateEntityFrom = (args: NonPersistedXStateStoreEvent): StateEntity => {
+  const stateEntity = new StateEntity()
+  stateEntity.step = args.step
```
nklomp, line 39: Similar comment as for the other PR. This is prone to future bugs by missing a mapping, whilst we simply can do a shallow copy

packages/data-store/src/xstate/XStateStore.ts
```diff
@@ -0,0 +1,86 @@
...
+    this.dbConnection = dbConnection
+  }
+
+  async saveState(state: SaveStateArgs): Promise<State> {
+    const connection: DataSource = await this.dbConnection
+    debug(`Executing saveState with state: ${JSON.stringify(state)}`)
+    return connection.getRepository(StateEntity).save(state)
+  }
+
+  async getState(args: GetStateArgs): Promise<State> {
```
nklomp, line 25: I am not really sure about this method. It makes an assumption there will be only one state of a certain type. It does not take into account tenants at all and also does not look at expired states etc. IMO This method should return the most recent non-final Xstate instance of the type, but also look at the tenant and whether it is not final

nklomp (dismissed):
See remarks

packages/xstate-persistence/src/types/types.ts
sksadjad: yeah, it didn't make any sense so changed the logic of deletion

packages/xstate-persistence/.eslintrc.json
```diff
@@ -0,0 +1,50 @@
+{
```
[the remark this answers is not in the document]
sksadjad: removed it

packages/xstate-persistence/package.json
```diff
@@ -0,0 +1,50 @@
...
+  "scripts": {
+    "build": "tsc",
+    "build:clean": "tsc --build --clean && tsc --build",
+    "generate-plugin-schema": "ts-node ../../packages/dev/bin/sphereon.js dev generate-plugin-schema"
+  },
+  "devDependencies": {
+    "@sphereon/ssi-sdk.agent-config": "workspace:*",
+    "@sphereon/ssi-sdk.core": "workspace:*",
+    "@sphereon/ssi-sdk.data-store": "workspace:*",
+    "@types/node": "^20.11.17",
```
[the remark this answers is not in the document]
sksadjad: I double checked this similar modules that we have here and removed this dependency

packages/xstate-persistence/src/agent/XStatePersistence.ts
```diff
@@ -0,0 +1,73 @@
...
+      case XStatePersistenceEventType.EVERY:
+        // Calling the context of the agent to make sure the REST client is called when configured
+        await context.agent.persistMachineSnapshot({ ...event.data })
+        break
+      default:
+        return Promise.reject(Error('Event type not supported'))
+    }
+  }
+
+  private async persistMachineSnapshot(args: NonPersistedMachineSnapshot): Promise<State> {
```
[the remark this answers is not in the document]
sksadjad: changed all the method names to what you have described.

packages/xstate-persistence/tsconfig.json
```diff
@@ -0,0 +1,32 @@
...
+    "outDir": "dist",
+    "declarationDir": "dist",
+    "esModuleInterop": true,
+    "skipLibCheck": true,
+    "resolveJsonModule": true
+  },
+  "include": ["src/**/*.json", "src/**/*.ts"],
+  "references": [
+    {
+      "path": "../vc-status-list"
```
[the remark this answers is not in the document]
sksadjad: removed this and other unnecessary references

packages/data-store/src/entities/xstate/StateEntity.ts
sksadjad: changed this to an object assign

packages/data-store/src/xstate/XStateStore.ts
sksadjad: changed the way we're handling this. made this look into the `expiresAt` and `completedAt` fields and return the most recent one

packages/data-store/src/entities/xstate/StateEntity.ts
sksadjad: changed to `xStateEventType` with column name: `xstate_event_type`

packages/data-store/src/xstate/XStateStore.ts
sksadjad: changed to the suggested impl

packages/data-store/src/types/xstate/IAbstractXStateStore.ts
sksadjad: made the state in the entity `string` and in the plugin `unknown`

packages/data-store/src/xstate/XStateStore.ts
nklomp: Much better. Although I am still a bit about the fence of returning only one entity.  I think we better return all of them, as I do believe that in the future we can have multiple instances of a flow running at the same time (for instance multiple people from the same tenant). I rather see us return all results. Any user of the method then can always get the first result or do some additional matching in the results

packages/data-store/src/entities/xstate/StateEntity.ts
nklomp: Okay apparently the generated value is not unique. So we need something like this to ensure uniqueness. Will take care of that

packages/data-store/src/xstate/XStateStore.ts
nklomp: Will change this to return multiple results so the user can choose

packages/data-store/src/types/xstate/IAbstractXStateStore.ts
nklomp: In the current version the store is exposing a string, as that should know nothing about the implementation. The xstate-machine-persistence plugin has a simplified State interface that resembles the State object of xstate but is more generic

packages/data-store/src/xstate/XStateStore.ts
nklomp: It is now returning an array

packages/data-store/src/entities/xstate/StateEntity.ts
nklomp: Currently we have an id that can be passed in and is expected to be globally unique. Otherwise it gets generated.
There is also a sessionId, that takes the "unique" value from Xstate

nklomp:
Okay PR is ready for review. A lot has been changed. The most important part is that the store is agnostic, whilst the xstate-machine-persist plugin is Xstate specific. It accepts Xstate objects. This is different from the original implementation, where the burden would be constantly on the parties calling the plugin to provide all the properties and convert xstate objects to the desired formats. Since that is not a nice plugin from a user perspective it now has been changed to be specific for Xstate.

I still need to create one method for rehydration and continuing the machine interpreter from the persisted state in an easy way. But I will do that when working on the wallet authorization code flow

maikel-maas (commented):
LGTM, but I'm the author of the PR, I need someone else to approve it

nklomp (approved):
Merging

PR #163, opened 2024-02-21
feature/SDK-1

BtencateSphereon:


packages/oid4vci-holder/README.md
```diff
@@ -0,0 +1,27 @@
+<!--suppress HtmlDeprecatedAttribute -->
+<h1 align="center">
+  <br>
+  <a href="https://www.sphereon.com"><img src="https://sphereon.com/content/themes/sphereon/assets/img/logo.svg" alt="Sphereon" width="400"></a>
+  <br>OpenID for Verifiable Credentials Holder (OID4VCI)
```
nklomp, line 5:
```suggestion
  <br>OpenID for Verifiable Credentials Issuance - Holder (OID4VCI)
```

packages/oid4vci-holder/src/agent/OID4VCIHolder.ts
```diff
@@ -0,0 +1,289 @@
...
+  public async onEvent(event: any, context: RequiredContext): Promise<void> {
+    switch (event.type) {
+      case OID4VCIHolderEvent.CONTACT_IDENTITY_CREATED:
+        this.onContactIdentityCreated?.(event.data)
+        break
+      case OID4VCIHolderEvent.CREDENTIAL_STORED:
+        this.onCredentialStored?.(event.data)
+        break
+      default:
+        return Promise.reject(Error('Event type not supported'))
```
nklomp, line 91:
```suggestion
        return Promise.reject(Error(`Event type '${event.type}' not supported`))
```

packages/oid4vci-holder/src/agent/OID4VCIHolder.ts
```diff
@@ -0,0 +1,289 @@
...
+
+    const oid4vciMachineInstanceArgs: OID4VCIMachineInstanceOpts = {
+      ...args,
+      services: {
+        ...services,
+        ...args.services,
+      },
+    }
+
+    return OID4VCIMachine.newInstance(oid4vciMachineInstanceArgs)
```
nklomp, line 115: Could we also return a unique instance Id, or get that from the machine. So that we return an object. One being the interpreter and one being the id of the instance and for instance something like the name of the machine? That will be helpful probably for the (re)hydration support we are also creating

packages/oid4vci-holder/src/agent/OID4VCIHolder.ts
```diff
@@ -0,0 +1,289 @@
...
+
+    if (requestData?.uri === undefined) {
+      return Promise.reject(Error('Missing request uri in context'))
+    }
+
+    if (
+      !requestData?.uri ||
+      !(requestData?.uri.startsWith(RequestType.OPENID_INITIATE_ISSUANCE) || requestData?.uri.startsWith(RequestType.OPENID_CREDENTIAL_OFFER))
+    ) {
+      return Promise.reject(Error('Invalid Uri'))
```
nklomp, line 129: Please provide a bit more context, If you want to keep messages to a user clean we will need to device  an error probably that can have more context for logging and a simple message for users. Or we do the logging here, but that is a bit of an anti-pattern.

In this case it would be very helpful to have the actual URI in the logs in case it goes wrong

nklomp (changes requested):
Some small remarks. I think it would be good if we have some machine identifier when creating a new interpreter.

packages/oid4vci-holder/README.md
BtencateSphereon: changed

packages/oid4vci-holder/src/agent/OID4VCIHolder.ts
BtencateSphereon: changed

packages/oid4vci-holder/src/agent/OID4VCIHolder.ts
BtencateSphereon: refactored, there is no machine name. so no sure what you want to do here

packages/oid4vci-holder/src/agent/OID4VCIHolder.ts
BtencateSphereon: added the uri

PR #165, opened 2024-02-22
Feature/CWALL-130 - tenant aware credential store

sksadjad:


packages/data-store/src/types/uniformCredential/IAbstractUniformCredentialStore.ts
```diff
@@ -0,0 +1,29 @@
...
+  tenantId?: string
+  expiresAt?: Date
+  state?: CredentialStateType
+  verificationDate?: Date
+  revocationDate?: Date
+}
+
+export type UpdateUniformCredentialStateArgs = GetUniformCredentialArgs & { verifiedState: CredentialStateType }
+
+export type RemoveUniformCredentialArgs = { id: string }
```
nklomp, line 29: Why does get have id or hash and Remove only id?

packages/data-store/src/uniformCredential/AbstractUniformCredentialStore.ts
```diff
@@ -0,0 +1,16 @@
...
+  GetUniformCredentialArgs,
+  GetUniformCredentialsArgs,
+  RemoveUniformCredentialArgs,
+  UpdateUniformCredentialStateArgs,
+} from '../types/uniformCredential/IAbstractUniformCredentialStore'
+import { UniformCredentialEntity } from '../entities/uniformCredential/UniformCredentialEntity'
+
+export abstract class AbstractUniformCredentialStore {
+  abstract getUniformCredential(args: GetUniformCredentialArgs): Promise<UniformCredentialEntity>
+  abstract getUniformCredentials(args?: GetUniformCredentialsArgs): Promise<Array<UniformCredentialEntity>>
```
nklomp, line 12: We need to add pagination support, given we know we will need this

packages/data-store/src/migrations/sqlite/1708525189002-CreateUniformCredential.ts
```diff
@@ -0,0 +1,33 @@
...
+
+export class CreateUniformCredential1708525189002 implements MigrationInterface {
+  name = 'CreateUniformCredential1708525189002'
+
+  public async up(queryRunner: QueryRunner): Promise<void> {
+    await queryRunner.query(`
+            CREATE TABLE "UniformCredential" (
+                "id" varchar PRIMARY KEY NOT NULL,
+                "credential_type" varchar CHECK( "credential_type" IN ('vc', 'vp') ) NOT NULL,
+                "document_format" varchar CHECK( "document_format" IN ('JSON-LD', 'JWT', 'SD-JWT', 'MDOC') ) NOT NULL,
```
nklomp, line 11: Consistency. Either have the enums in caps or lowercase, but not one in upper and others in lower. In order to prevent issues with mapping DTO's either user string literals instead of enums, or prevent things like dashes and have the enum in the same case as the value.
Typeorm has enum support, but be aware that the DTOs cross from frontend to backend, so if you re-use the same enums between entities and DTO's you are now in trouble

packages/data-store/src/types/uniformCredential/IAbstractUniformCredentialStore.ts
```diff
@@ -0,0 +1,29 @@
+import { CredentialCorrelationType, CredentialDocumentFormat, CredentialStateType, CredentialTypeEnum, UniformCredential } from './uniformCredential'
+
+export type GetUniformCredentialArgs = { id: string } | { hash: string }
```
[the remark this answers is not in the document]
sksadjad: refactored to `*DigitalCredentialArgs`

packages/data-store/src/types/uniformCredential/uniformCredential.ts
```diff
@@ -0,0 +1,41 @@
...
+  subjectCorrelationId?: string
+  uniformDocument: string
+  lastVerifiedState?: CredentialStateType
+  tenantId?: string
+  createdAt: Date
+  lastUpdatedAt: Date
+  expiresAt?: Date
+}
+
+export enum CredentialTypeEnum {
```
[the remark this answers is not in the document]
sksadjad: removed

packages/data-store/src/uniformCredential/UniformCredentialStore.ts
```diff
@@ -0,0 +1,97 @@
...
+  updateUniformCredentialState = async (args: UpdateUniformCredentialStateArgs): Promise<UniformCredentialEntity> => {
+    const credentialRepository: Repository<UniformCredentialEntity> = (await this.dbConnection).getRepository(UniformCredentialEntity)
+    const whereClause: Record<string, any> = {}
+    if ('id' in args) {
+      whereClause.id = args.id
+    }
+    if ('hash' in args) {
+      whereClause.hash = args.hash
+    }
+    const credential: UniformCredentialEntity | null = await credentialRepository.findOne({
```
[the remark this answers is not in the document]
sksadjad: added a check and if `id` or `hash` is not provided, returning `false`

packages/data-store/src/uniformCredential/UniformCredentialStore.ts
```diff
@@ -0,0 +1,97 @@
...
+    let error = false
+    const result = await (
+      await this.dbConnection
+    )
+      .getRepository(UniformCredentialEntity)
+      .delete({
+        id: args.id,
+      })
+      .catch((error) => {
+        error = true
```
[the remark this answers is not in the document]
sksadjad: refactored this part and just returning false in case of error (or in the case of `result.affected!==1` )

packages/data-store/src/types/uniformCredential/uniformCredential.ts
```diff
@@ -0,0 +1,41 @@
+export type NonPersistedUniformCredential = Omit<UniformCredential, 'id'>
+
+export type UniformCredential = {
```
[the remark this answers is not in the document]
sksadjad: changed to DigitalCredential

packages/data-store/src/uniformCredential/UniformCredentialStore.ts
```diff
@@ -0,0 +1,97 @@
...
+const debug: Debug.Debugger = Debug('sphereon:ssi-sdk:credential-store')
+export class UniformCredentialStore extends AbstractUniformCredentialStore {
+  private readonly dbConnection: OrPromise<DataSource>
+
+  constructor(dbConnection: OrPromise<DataSource>) {
+    super()
+    this.dbConnection = dbConnection
+  }
+
+  addUniformCredential = async (args: AddUniformCredentialArgs): Promise<UniformCredentialEntity> => {
```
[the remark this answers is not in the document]
sksadjad: changed the name of the functions to `(operation)DigitalCredential`

packages/data-store/src/utils/uniformCredential/MappingUtils.ts
```diff
@@ -0,0 +1,75 @@
...
+import { computeEntryHash } from '@veramo/utils'
+
+export const uniformCredentialEntityFromAddArgs = (addCredentialArgs: AddUniformCredentialArgs): UniformCredentialEntity => {
+  const wrappedCredential: WrappedVerifiableCredential | WrappedVerifiablePresentation =
+    addCredentialArgs.credentialType === CredentialTypeEnum.VC
+      ? CredentialMapper.toWrappedVerifiableCredential(JSON.stringify(addCredentialArgs.raw))
+      : CredentialMapper.toWrappedVerifiablePresentation(addCredentialArgs.raw)
+  const uniformCredentialEntity: UniformCredentialEntity = new UniformCredentialEntity()
+
+  uniformCredentialEntity.credentialType = addCredentialArgs.credentialType
```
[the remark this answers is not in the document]
sksadjad: changed it with a copy

packages/data-store/src/utils/uniformCredential/MappingUtils.ts
```diff
@@ -0,0 +1,75 @@
+import { AddUniformCredentialArgs } from '../../types/uniformCredential/IAbstractUniformCredentialStore'
+import { UniformCredentialEntity } from '../../entities/uniformCredential/UniformCredentialEntity'
+import { CredentialMapper, WrappedVerifiableCredential, WrappedVerifiablePresentation } from '@sphereon/ssi-types'
+import { CredentialTypeEnum, NonPersistedUniformCredential, UniformCredential } from '../../types/uniformCredential/uniformCredential'
+import { computeEntryHash } from '@veramo/utils'
+
+export const uniformCredentialEntityFromAddArgs = (addCredentialArgs: AddUniformCredentialArgs): UniformCredentialEntity => {
+  const wrappedCredential: WrappedVerifiableCredential | WrappedVerifiablePresentation =
```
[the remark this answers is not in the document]
sksadjad: changed to use uniform methods

packages/data-store/src/utils/uniformCredential/MappingUtils.ts
```diff
@@ -0,0 +1,75 @@
...
+  uniformCredentialEntity.uniformDocument =
+    addCredentialArgs.credentialType === CredentialTypeEnum.VC
+      ? JSON.stringify((wrappedCredential as WrappedVerifiableCredential).credential)
+      : JSON.stringify((wrappedCredential as WrappedVerifiablePresentation).presentation)
+  return uniformCredentialEntity
+}
+
+export const uniformCredentialEntityFromNonPersisted = (uniformCredential: NonPersistedUniformCredential): UniformCredentialEntity => {
+  const uniformCredentialEntity: UniformCredentialEntity = new UniformCredentialEntity()
+  uniformCredentialEntity.credentialType = uniformCredential.credentialType
```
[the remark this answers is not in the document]
sksadjad: changed to copy

packages/data-store/src/types/uniformCredential/IAbstractUniformCredentialStore.ts
sksadjad: added `hash` to parameters of delete

packages/data-store/src/migrations/sqlite/1708525189002-CreateUniformCredential.ts
sksadjad: changed the enums to havesame name and value, and second made them all caps

packages/data-store/src/utils/uniformCredential/MappingUtils.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export const uniformCredentialEntityFromAddArgs = (addCredentialArgs: AddUniformCredentialArgs): UniformCredentialEntity => {
+  const wrappedCredential: WrappedVerifiableCredential | WrappedVerifiablePresentation =
+    addCredentialArgs.credentialType === CredentialTypeEnum.VC
+      ? CredentialMapper.toWrappedVerifiableCredential(JSON.stringify(addCredentialArgs.raw))
+      : CredentialMapper.toWrappedVerifiablePresentation(addCredentialArgs.raw)
+  const uniformCredentialEntity: UniformCredentialEntity = new UniformCredentialEntity()
+
+  uniformCredentialEntity.credentialType = addCredentialArgs.credentialType
+  uniformCredentialEntity.createdAt = new Date()
```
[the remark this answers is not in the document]
sksadjad: added `issuedAt` and added a function to find it from the raw document

packages/data-store/src/utils/uniformCredential/MappingUtils.ts
```diff
@@ -0,0 +1,75 @@
...
+  uniformCredentialEntity.lastUpdatedAt = new Date()
+  uniformCredentialEntity.tenantId = addCredentialArgs.tenantId
+  uniformCredentialEntity.raw = addCredentialArgs.raw
+  uniformCredentialEntity.issuerCorrelationId = addCredentialArgs.issuerCorrelationId
+  uniformCredentialEntity.issuerCorrelationType = addCredentialArgs.issuerCorrelationType
+  uniformCredentialEntity.subjectCorrelationId = addCredentialArgs.subjectCorrelationId
+  uniformCredentialEntity.subjectCorrelationType = addCredentialArgs.subjectCorrelationType
+  uniformCredentialEntity.expiresAt = addCredentialArgs.expiresAt
+  uniformCredentialEntity.lastVerifiedState = addCredentialArgs.state
+  uniformCredentialEntity.verificationDate = addCredentialArgs.verificationDate
```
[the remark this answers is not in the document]
sksadjad: refactored to `lastVerificationDate`

packages/data-store/src/types/uniformCredential/IAbstractUniformCredentialStore.ts
```diff
@@ -0,0 +1,29 @@
...
+export type AddUniformCredentialArgs = {
+  credentialType: CredentialTypeEnum
+  documentFormat: CredentialDocumentFormat
+  raw: string
+  issuerCorrelationType: CredentialCorrelationType
+  subjectCorrelationType?: CredentialCorrelationType
+  issuerCorrelationId: string
+  subjectCorrelationId?: string
+  tenantId?: string
+  expiresAt?: Date
```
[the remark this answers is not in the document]
sksadjad: removed this parameter and added a function to get it from the raw data

packages/data-store/src/types/uniformCredential/IAbstractUniformCredentialStore.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export type FindUniformCredentialArgs = Array<Partial<UniformCredential>>
+
+// TODO: discuss about what args we want here
+export type GetUniformCredentialsArgs = {
+  filter?: FindUniformCredentialArgs
+}
+
+export type AddUniformCredentialArgs = {
+  credentialType: CredentialTypeEnum
```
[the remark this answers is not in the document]
sksadjad: removed this and added a function to get it from the raw data

packages/data-store/src/uniformCredential/AbstractUniformCredentialStore.ts
sksadjad: added pagination and sorting to the parameters of get many method

packages/data-store/src/types/uniformCredential/IAbstractUniformCredentialStore.ts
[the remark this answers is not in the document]
nklomp: Okay that wasn't exactly what I hoped you would do. If I am already accessing the DigitalCredentialStore, I do not have to name every single method DigitalCredential (that is already implied for a Credential Store with the name Digital in it.

What you can end up is someone calling the store instance constant `digitalCredentialStore` and then you thus end up with `digitialCredentialStore.getDigitalCredential()`. Just call these methods `xxxCredential`. They are fully typed, thus it will be clear it either accepts or returns Digital Credentials. The word digital adds nothing in all the method names

packages/data-store/src/types/uniformCredential/IAbstractUniformCredentialStore.ts
```diff
@@ -0,0 +1,29 @@
...
+  documentFormat: CredentialDocumentFormat
+  raw: string
+  issuerCorrelationType: CredentialCorrelationType
+  subjectCorrelationType?: CredentialCorrelationType
+  issuerCorrelationId: string
+  subjectCorrelationId?: string
+  tenantId?: string
+  expiresAt?: Date
+  state?: CredentialStateType
```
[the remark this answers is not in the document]
nklomp: I mentioned verificationState for a reason, because the other related property is called verificationDate (which btw is a poor name, because for verification the actual time is also pretty important, contrary to issuance and revocation. For the latter there of course also is a time associated but these are typically important at a day granularity. Having said that, for VCDM 2.0 the names have moved to validFrom and validUntil, so it is probably best to already use these for our DB columns as well. https://www.w3.org/TR/vc-data-model-2.0/#validity-period

So either call both verification of verified, but do not mix them. So suggest to call it `verifiedState` and `verifiedAt` or something. One is about when it happened, the other is the outcome at that specific point in time. We mainly need these for fast and easy access in the UI

packages/data-store/src/digitalCredential/DigitalCredentialStore.ts
```diff
@@ -0,0 +1,116 @@
...
+  updateDigitalCredentialState = async (args: UpdateDigitalCredentialStateArgs): Promise<DigitalCredentialEntity> => {
+    const credentialRepository: Repository<DigitalCredentialEntity> = (await this.dbConnection).getRepository(DigitalCredentialEntity)
+    const whereClause: Record<string, any> = {}
+    if ('id' in args) {
+      whereClause.id = args.id
+    }
+    if ('hash' in args) {
+      whereClause.hash = args.hash
+    }
+    const credential: DigitalCredentialEntity | null = await credentialRepository.findOne({
```
nklomp, line 98: As mentioned elsewhere. Please check the where clause itself. It either has to contain an id or hash. Don't use the persistence Layer to figure that out. You are now relying on the persistence layer, whilst you can do a check up front. For instance what happens with a findOne if there is no where clause? Would it select the first result of all entities, or would it give an error? Just prevent that altogether

packages/data-store/src/types/digitalCredential/IAbstractDigitalCredentialStore.ts
```diff
@@ -0,0 +1,34 @@
+import { CredentialCorrelationType, CredentialStateType, DigitalCredential } from './digitalCredential'
+
+export type GetDigitalCredentialArgs = { id: string } | { hash: string }
+
+export type FindDigitalCredentialArgs = Array<Partial<DigitalCredential>>
+
+export type GetDigitalCredentialsArgs = {
+  filter?: FindDigitalCredentialArgs
+  skip?: number
+  take?: number
```
nklomp, line 10: Did you doublecheck this with how the cloud wallet expects pagination to happen?

packages/data-store/src/utils/digitalCredential/MappingUtils.ts
```diff
@@ -0,0 +1,112 @@
...
+  CredentialDocumentFormat,
+  CredentialType,
+  DigitalCredential,
+  NonPersistedDigitalCredential,
+} from '../../types/digitalCredential/digitalCredential'
+import { computeEntryHash } from '@veramo/utils'
+import { createHash } from 'crypto'
+
+function determineCredentialType(raw: string): CredentialType {
+  if (CredentialMapper.hasProof(raw)) {
```
nklomp, line 21: Hmzz, did you check this will work. Because you are now passing in a string, which is only a string for the persistence layer. So even a JSON-LD credential is a string at this point. The CredentialMapper of course can work with strings, but it might assume we are talking about JWTs when it is a string. In other words, probably at this point you should still be accepting the original/raw credential

packages/data-store/src/utils/digitalCredential/MappingUtils.ts
```diff
@@ -0,0 +1,112 @@
...
+    case DocumentFormat.JWT:
+      return CredentialDocumentFormat.JWT
+    case DocumentFormat.SD_JWT_VC:
+      return CredentialDocumentFormat.SD_JWT
+    default:
+      throw new Error(`Not supported document format: ${documentFormat}`)
+  }
+}
+
+function getExpiresAt(uniformDocument: IVerifiableCredential | IVerifiablePresentation | SdJwtDecodedVerifiableCredentialPayload): Date | undefined {
```
nklomp, line 50: See remark elsewhere about new names in VCDM 2; suggest to start using these

packages/data-store/src/utils/digitalCredential/MappingUtils.ts
```diff
@@ -0,0 +1,112 @@
...
+    return new Date(uniformDocument.issuanceDate)
+  } else if (uniformDocument.proof && CredentialMapper.getFirstProof(uniformDocument as IVerifiableCredential | IVerifiablePresentation)) {
+    return new Date(CredentialMapper.getFirstProof(uniformDocument as IVerifiableCredential | IVerifiablePresentation)!.created)
+  } else if ('validFrom' in uniformDocument) {
+    return new Date(uniformDocument['validFrom'])
+  } else if ('iat' in uniformDocument) {
+    return new Date(uniformDocument['iat'] * 1000)
+  }
+
+  return undefined
```
nklomp, line 70: Issuance date can never be undefined and thus should throw an error

packages/data-store/src/utils/digitalCredential/MappingUtils.ts
```diff
@@ -0,0 +1,112 @@
...
+function getExpiresAt(uniformDocument: IVerifiableCredential | IVerifiablePresentation | SdJwtDecodedVerifiableCredentialPayload): Date | undefined {
+  if ('expirationDate' in uniformDocument) {
+    return new Date(uniformDocument.expirationDate)
+  } else if ('exp' in uniformDocument) {
+    return new Date(uniformDocument.exp)
+  }
+  return undefined
+}
+
+function getIssuedAt(uniformDocument: IVerifiableCredential | IVerifiablePresentation | SdJwtDecodedVerifiableCredentialPayload): Date | undefined {
```
nklomp, line 59: See remark about new names in VCDM 2

packages/data-store/src/utils/digitalCredential/MappingUtils.ts
```diff
@@ -0,0 +1,112 @@
...
+  if ('issuanceDate' in uniformDocument) {
+    return new Date(uniformDocument.issuanceDate)
+  } else if (uniformDocument.proof && CredentialMapper.getFirstProof(uniformDocument as IVerifiableCredential | IVerifiablePresentation)) {
+    return new Date(CredentialMapper.getFirstProof(uniformDocument as IVerifiableCredential | IVerifiablePresentation)!.created)
+  } else if ('validFrom' in uniformDocument) {
+    return new Date(uniformDocument['validFrom'])
+  } else if ('iat' in uniformDocument) {
```
nklomp, line 66: If not mistaken, this should not be `iat` but `nbf`. Please doublecheck that in VCDM 1.1. If that is the case, suggest to add an else if with `nbf` in front of the `iat`

packages/data-store/src/utils/digitalCredential/MappingUtils.ts
```diff
@@ -0,0 +1,112 @@
...
+  SdJwtDecodedVerifiableCredentialPayload,
+} from '@sphereon/ssi-types'
+import {
+  CredentialDocumentFormat,
+  CredentialType,
+  DigitalCredential,
+  NonPersistedDigitalCredential,
+} from '../../types/digitalCredential/digitalCredential'
+import { computeEntryHash } from '@veramo/utils'
+import { createHash } from 'crypto'
```
nklomp, line 18: Please remove the createHash import

packages/data-store/src/utils/digitalCredential/MappingUtils.ts
```diff
@@ -0,0 +1,112 @@
...
+    return new Date(uniformDocument['iat'] * 1000)
+  }
+
+  return undefined
+}
+
+export const handleSdJwt = (rawCredential: string): SdJwtDecodedVerifiableCredentialPayload => {
+  // todo ask about the hasher
+  return CredentialMapper.isSdJwtEncoded(rawCredential)
+    ? decodeSdJwtVc(rawCredential, (data, algorithm) => createHash(algorithm).update(data).digest()).decodedPayload
```
nklomp, line 76: Please remove the createHash call. The hasher implementation needs to be passed in and called

packages/data-store/src/utils/digitalCredential/MappingUtils.ts
```diff
@@ -0,0 +1,112 @@
...
+export const nonPersistedDigitalCredentialEntityFromAddArgs = (addCredentialArgs: AddDigitalCredentialArgs): NonPersistedDigitalCredential => {
+  const credentialType: CredentialType = determineCredentialType(addCredentialArgs.raw)
+  const documentFormat: DocumentFormat = CredentialMapper.detectDocumentType(addCredentialArgs.raw)
+  const uniformDocument =
+    documentFormat === DocumentFormat.SD_JWT_VC
+      ? handleSdJwt(addCredentialArgs.raw)
+      : credentialType === CredentialType.VC || credentialType === CredentialType.C
+      ? CredentialMapper.toUniformCredential(addCredentialArgs.raw)
+      : CredentialMapper.toUniformPresentation(addCredentialArgs.raw)
+  const issuedAt: Date | undefined = getIssuedAt(uniformDocument)
```
nklomp, line 89: See remarks about new names used in VCDM2

packages/ssi-types/src/mapper/credential-mapper.ts
```diff
@@ -293,6 +293,82 @@ export class CredentialMapper {
...
 
+  public static isCredential(original: OriginalVerifiableCredential | OriginalVerifiablePresentation): boolean {
+    let document = original
+    try {
+      if (typeof original === 'string') {
+        if (this.isJwtEncoded(original)) {
+          const vc: IVerifiableCredential = this.toUniformCredential(original)
+          return this.isW3cCredential(vc)
+        } else if (this.isSdJwtEncoded(original)) {
+          // find out how we can differentiate between vc and vp for sd-jwts
```
nklomp, line 304: Remove this comment, because they are always a credential

packages/ssi-types/src/mapper/credential-mapper.ts
```diff
@@ -293,6 +293,82 @@ export class CredentialMapper {
...
+      )
+    } catch (e) {
+      return false
+    }
+  }
+
+  public static hasProof(original: OriginalVerifiableCredential | OriginalVerifiablePresentation | string): boolean {
+    let document = original
+    try {
+      if (this.isJwtEncoded(original) || this.isJwtDecodedCredential(original as OriginalVerifiableCredential)) {
```
nklomp, line 347: Hmzz, this behaves different from the getFirstProof method, which is now made public. The latter returns undefined for a JWT, suggesting there is no proof (that of course is debatable, because we really should look at whether a signature is present in the jwt (simply split the jwt by '.' and look at whether the 3rd index is present.

Whether that should be added to the getFirstProof method, or whether this method should behave the same as getFirstProof is not something I can answer from this quick review, because I need to know the context from which the getFirstProof method was called in the past.

I am making a fuss about it, because we now have 2 public methods related to proofs. One called hasProof which for a JWT returns true, but then the getFirstProof returns undefined for that same JWT. Simply adjusting the latter function might have undesired effects on existing code that called the previous internal function.

packages/ssi-types/src/mapper/credential-mapper.ts
```diff
@@ -293,6 +293,82 @@ export class CredentialMapper {
...
+
+  public static isPresentation(original: OriginalVerifiableCredential | OriginalVerifiablePresentation): boolean {
+    let document = original
+    try {
+      if (typeof original === 'string')
+        if (this.isJwtEncoded(original)) {
+          const vp: IVerifiablePresentation = this.toUniformPresentation(original)
+          return this.isW3cPresentation(vp)
+        } else if (this.isSdJwtEncoded(original)) {
+          return true
```
nklomp, line 329: Hmzz, this needs to be documented. Because an SD JWT VC is used as a presentation as well. Right now both the isCredential and isPresentation methods return true for the same passed in SD JWT. If the context of calling this method is to make a distinction between VCs and VPs then the isPresentation should return false. If however the context is to determine from code whether we are constructing or using a VP somewhere in the code, that it would be okay to return true here, because although SD JWTs are always VCs, they are also used as presentations.

My suggestion is to change this one to false. And then to adjust the logic in where the isPresentation and isCredential is currently called. If it needs a presentation somewhere, then you can just use true in case it is a SD-JWT

packages/ssi-types/src/mapper/credential-mapper.ts
```diff
@@ -293,6 +293,82 @@ export class CredentialMapper {
...
+        } else {
+          document = JSON.parse(original)
```
nklomp, line 331: This is a really confusing construct. You are constantly returning in the above branches, but now you have an else, and then outside of the else you start returning again. Just remove the else, or return from that else

packages/ssi-types/src/mapper/credential-mapper.ts
```diff
@@ -293,6 +293,82 @@ export class CredentialMapper {
...
+    try {
+      if (typeof original === 'string') {
+        if (this.isJwtEncoded(original)) {
+          const vc: IVerifiableCredential = this.toUniformCredential(original)
+          return this.isW3cCredential(vc)
+        } else if (this.isSdJwtEncoded(original)) {
+          // find out how we can differentiate between vc and vp for sd-jwts
+          return true
+        } else {
+          document = JSON.parse(original)
```
nklomp, line 307: This is a really confusing construct. You are constantly returning in the above branches, but now you have an else, and then outside of the else you start returning again. Just remove the else, or return from that else

packages/data-store/src/uniformCredential/UniformCredentialStore.ts
```diff
@@ -0,0 +1,97 @@
...
+      where: whereClause,
+    })
+
+    if (!credential) {
+      return Promise.reject(Error(`No credential found for args: ${whereClause}`))
+    }
+
+    const updatedCredential = {
+      ...credential,
+      lastUpdatedAt: new Date(),
```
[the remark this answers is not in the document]
sksadjad: passed an optional date here

packages/data-store/src/types/uniformCredential/IAbstractUniformCredentialStore.ts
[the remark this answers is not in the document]
sksadjad: changed the `lastVerificationDate` to `verifiedAt` also did that for `revocationDate` and changed it to `revokedAt`

packages/data-store/src/types/uniformCredential/IAbstractUniformCredentialStore.ts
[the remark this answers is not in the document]
sksadjad: removed the digital keywords from the function names. now we have: `getCredential
getCredentials
addCredential
updateCredentialState
removeCredential`

packages/data-store/src/digitalCredential/DigitalCredentialStore.ts
```diff
@@ -0,0 +1,116 @@
...
+    })
+
+    if (!credential) {
+      return Promise.reject(Error(`No credential found for args: ${whereClause}`))
+    }
+    const updatedCredential: DigitalCredential = {
+      ...credential,
+      lastVerificationDate: args.verificationDate ?? new Date(),
+      lastUpdatedAt: new Date(),
+      ...(args.verifiedState === CredentialStateType.REVOKED && { revocationDate: args.verificationDate ?? new Date() }),
```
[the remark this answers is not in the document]
sksadjad: added a new property to the args of this param and added throw error if it's not provided in case of revoke

packages/data-store/src/digitalCredential/DigitalCredentialStore.ts
```diff
@@ -0,0 +1,116 @@
...
+    const credential: DigitalCredentialEntity | null = await credentialRepository.findOne({
+      where: whereClause,
+    })
+
+    if (!credential) {
+      return Promise.reject(Error(`No credential found for args: ${whereClause}`))
+    }
+    const updatedCredential: DigitalCredential = {
+      ...credential,
+      lastVerificationDate: args.verificationDate ?? new Date(),
```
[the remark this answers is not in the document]
sksadjad: not sure if I understand this comment. this is updating the state of a stored credential and not adding a new credential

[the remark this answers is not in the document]
nklomp: The adding part wasn't the point. Read 'updating'. **"automatically result in a lastVerificationDate if not provided"** The next sentence should also be pretty clear what the issue is with the way you are setting the lastVerificationDate.
I suggest you reread the remarks, read the code setting the lastVerificationDate and then tell me in your own words, what is wrong with it, because it should be immediately clear to you if you ask me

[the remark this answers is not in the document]
sksadjad: if I understand correctly, we should always get a date in the params of the `updateCredentialState` it's `revokedAt` for when it's revoked and `verifiedAt` when it's to another state. and never should set it with current date.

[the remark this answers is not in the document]
nklomp: Yes, but also only set or overwrite it if the verification date is passed in. It is optional, because I could just as well be storing a VC/VP without verifying, or just playnly storing a C/P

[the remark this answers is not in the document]
nklomp: The most important thing of course was that persisting a (V)C/(V)P and providing some properties to make it more easily accessible for indexing/UX etc, has 0 relation with the store determining by itself when the last verification apparently happened if it was not passed in as argument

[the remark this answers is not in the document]
sksadjad: I hope I got it right this time. no setting the `verifiedAt` date when it's not provided. `verifiedAt` is optional in the entity itself.

packages/data-store/src/types/digitalCredential/IAbstractDigitalCredentialStore.ts
sksadjad: updated this impl to be more inline with supabase client handling of sorting/pagination

packages/data-store/src/utils/digitalCredential/MappingUtils.ts
sksadjad: yes I have many test cases for this. but because of your other [comment](https://github.com/Sphereon-Opensource/SSI-SDK/pull/165#discussion_r1506800014) I'll pass in the parsed version to the CredentialMapper. that way I don't need to have additional checks for object as string vs jwt/sd_jwt strings

sksadjad: changed this to pass in only a jwt/sd_jwt or json document to the CredentialMapper functions

packages/ssi-types/src/mapper/credential-mapper.ts
sksadjad: removed

packages/ssi-types/src/mapper/credential-mapper.ts
sksadjad: I personally think the impl is wrong on the getFirstProof method. jwt encoded credentials have been signed and we can get the signature the way you have described. but exactly because of your last point I didn't change the existing impl. I'm also confused about the `decodeVerifiableCredential` which is creating an embedded proof for the decoded credential and passing in the whole document in the jwt part of the proof.

packages/ssi-types/src/mapper/credential-mapper.ts
sksadjad: changed this to `false` for now, but we have to have a discussion about `sd_jwt`s and how are we going to support these. right now it's partially supported and in some cases it will lead to unwanted results.

packages/ssi-types/src/mapper/credential-mapper.ts
sksadjad: changed this structure. now every function in our credential-mapper will receive a jwt string/sd_jwt string/ json document

packages/ssi-types/src/mapper/credential-mapper.ts
sksadjad: changed this structure. now every function in our credential-mapper will receive a jwt string/sd_jwt string/ json document

packages/ssi-types/src/mapper/credential-mapper.ts
nklomp: Without looking at it I already tend to disagree about it being wrong. Why? The method was private for a reason and thus is constrained to a very specific use-case. That use-case is probably either not involving JWTs or has some other detections going on. The word Proof can be interpreted on that method in 2 ways. One is that a signature should be present (which is how you are now using it), the other simply is litteraly about the Proof object being present in the VC or not. 

The problem happens because you made the method public, and then of course the expectations could be completely different. So you first need to look at the code path(s) calling the previously private method, to determine whether the method was wrong to begin with or not; whether it assumed a signature or proof object. Only then you can decide whether the method needs changes or not when it needs to become public.

Regarding the decoding and passing a jwt part. That is literally how we and Veramo create a uniform representation, transforming a JWT string into a json ld like credential, allowing access to the claims. Then by adding a constructed proof, with value JwtProof2020 and passing in the full JWT VC, you have a Proof method that can be verified using the original JWT and you have a VC that can easily be used just like any other JSONLD VC

packages/data-store/src/utils/digitalCredential/MappingUtils.ts
sksadjad: for the issuance date, I'm also checking the `validFrom` from VCDM2 here. Altough the wording of the document is quite confusing by itself:
`If validFrom and validUntil are not present, the [verifiable credential](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verifiable-credential) validity period is considered valid indefinitely. In such cases, the [verifiable credential](https://www.w3.org/TR/vc-data-model-2.0/#dfn-verifiable-credential) is assumed to be valid from the time the verifiable credential was created.`
which suggests it's different from creation of the verifiable credential. We also have this:
from [here](https://www.w3.org/TR/vc-data-model-2.0/#validity-period)
`... which could be a date and time in the future or in the past.`
(from the changes from version 1):
- Make validFrom optional.
- Rename issuanceDate/expirationDate to validFrom/validUntil.

my suggestion is we move to signatures for this matter and use that instead. because every verifiable claim is signed and we can work better with them. Of course we can have extra fields in the db representing the new changes in the VCDM (for validFrom) but since it's optional in the data model, we can't rely on that for our VCs

packages/data-store/src/utils/digitalCredential/MappingUtils.ts
sksadjad: yes. in the VCDM 1 they are mentioning `nbf`. added `nbf` right before `iat`

packages/data-store/src/utils/digitalCredential/MappingUtils.ts
sksadjad: this can be undefined because we can pass unsigned documents to it.

packages/data-store/src/utils/digitalCredential/MappingUtils.ts
sksadjad: and by the way this is why I've made the getFirstProof method public to use it in here and get the signature's created date.

packages/data-store/src/utils/digitalCredential/MappingUtils.ts
sksadjad: removed and added it to the callback params

packages/data-store/src/utils/digitalCredential/MappingUtils.ts
sksadjad: removed and added it to the callback params

packages/data-store/src/utils/digitalCredential/MappingUtils.ts
sksadjad: changed this to validFrom

packages/data-store/src/utils/digitalCredential/MappingUtils.ts
sksadjad: also changed this to `validFrom`

packages/data-store/src/utils/digitalCredential/MappingUtils.ts
sksadjad: changed it to `validFrom`

packages/ssi-types/src/mapper/credential-mapper.ts
sksadjad: I reverted and made the `getFirstProof` private again. I'm using the functions `toUniformCredential` and `toUniformPresentation` functions in my mapper, and it's using this function and later I'm getting the proof value from the resulted `IVerifiableCredential`/`IVerifiablePresentation`

packages/data-store/src/utils/digitalCredential/MappingUtils.ts
nklomp: Signing has nothing to do with validFrom (previous Issuance)dates/times

packages/data-store/src/utils/digitalCredential/MappingUtils.ts
nklomp: Signing has nothing to do with validFrom (previous Issuance)dates/times

packages/data-store/src/utils/digitalCredential/MappingUtils.ts
sksadjad: issuanceDate (validFrom) is optional in VCDM 2. but does your comment mean that I can't use the signature date as issuanceDate? if so I have to alter my function for getting that

sksadjad: issuanceDate (validFrom) is optional in VCDM 2.

nklomp: No you shouldn't be using that. VCs can have multiple signatures. There is a reason for the name change to begin with

sksadjad: yeah removed that part of the logic

packages/data-store/src/types/digitalCredential/digitalCredential.ts
```diff
@@ -0,0 +1,46 @@
+export type NonPersistedDigitalCredential = Omit<DigitalCredential, 'id'>
+
+export type DigitalCredential = {
+  id: string
+  credentialType: CredentialType
```
nklomp, line 5: Suggest to rename this one to documentType, as we soon will also need the actual VC types in here for easy access, and then it become really confusing. Since a VP isn't a credential anyway and since documentFormat and uniformDocument are also already being used, suggest to stick with that

packages/data-store/src/types/digitalCredential/digitalCredential.ts
```diff
@@ -0,0 +1,46 @@
...
+  documentFormat: CredentialDocumentFormat
+  raw: string
```
nklomp, line 7: Suggest to rename this one to rawDocument, as we also already have uniformDocument

packages/data-store/src/digitalCredential/DigitalCredentialStore.ts
```diff
@@ -0,0 +1,124 @@
...
+
+    if (!result) {
+      return Promise.reject(Error(`No credential found for arg: ${args.toString()}`))
+    }
+    return result
+  }
+
+  getCredentials = async (args?: GetCredentialsArgs): Promise<GetCredentialsResponse> => {
+    const { filter = {}, offset, limit, order = 'id.asc' } = args ?? {}
+    const sortOptions: FindOptionsOrder<DigitalCredentialEntity> = parseAndValidateOrderOptions<DigitalCredentialEntity>(order)
```
nklomp, line 50: Hmzz, I would expect order to also support the result of the parseFunction as type. So that we parse whenever we get a string, but otherwise use the sortOptions as is. That would be the best of both worlds. Have fully typed sort options in case you are calling it from a service, but the option to use a string when coming from the frontend

packages/data-store/src/types/digitalCredential/IAbstractDigitalCredentialStore.ts
nklomp: See remark, as this is a different layer, it should not only behave directly like how supabase wants it. The point really was that our interface should be relative easy. For instance if we work with X and limit use that, instead of (from) X and (to) Y.

If the comment on the sort interface is solved we should be good for both

packages/data-store/src/types/digitalCredential/digitalCredential.ts
sksadjad: renamed to `documentType`

packages/data-store/src/types/digitalCredential/digitalCredential.ts
sksadjad: renamed to `rawDocument`

packages/data-store/src/digitalCredential/DigitalCredentialStore.ts
sksadjad: added `FindOptionsOrder<DigitalCredentialEntity>` to the parameters of get function

nklomp (approved):

PR #171, opened 2024-03-19
feature/CWALL-153

maikel-maas:


packages/data-store/src/entities/contact/StudentEntity.ts
```diff
@@ -0,0 +1,51 @@
+import {BaseContactEntity} from "./BaseContactEntity";
+import {BeforeInsert, BeforeUpdate, ChildEntity, Column} from "typeorm";
+import {IsNotEmpty, Validate, ValidationError, validate} from "class-validator";
+import {IsNonEmptyStringConstraint} from "../validators";
+import {ValidationConstraint} from "../../types";
```
nklomp, line 5: Why are we adding a Student Entity? That is not at all what we want. We discussed to be able to extend individuals with additional properties. We are not creating a product around students, so we should be flexible in supporting new individual types, without having to provide all explicit entities for these types.

packages/data-store/src/types/contact/contact.ts
```diff
@@ -153,6 +153,23 @@ export type NaturalPerson = {
 export type NonPersistedNaturalPerson = Omit<NaturalPerson, 'id' | 'createdAt' | 'lastUpdatedAt'>
 export type PartialNaturalPerson = Partial<NaturalPerson>
 
+export type Student = {
```
nklomp, line 156: Why are we adding a Student Entity? That is not at all what we want. We discussed to be able to extend individuals with additional properties. We are not creating a product around students, so we should be flexible in supporting new individual types, without having to provide all explicit entities for these types.

nklomp (changes requested):
We discussed to be able to extend contacts/individuals with metadata like on the party level. We are not creating a student product. To add student entities make no sense, and was not wat was discussed during sprint start

packages/data-store/src/entities/contact/StudentEntity.ts
sanderPostma: All instances of student are killed now

packages/contact-manager/__tests__/shared/contactManagerAgentLogic.ts
```diff
@@ -36,7 +37,7 @@ export default (testContext: { getAgent: () => ConfiguredAgent; setup: () => Pro
         displayName: 'default_display_name',
         contactType: {
           type: PartyTypeType.NATURAL_PERSON,
-          origin: PartyOrigin.EXTERNAL,
+          partyOrigin: PartyOrigin.EXTERNAL,
```
sanderPostma, line 40: origin fields were renamed because when the party & identity entities are joined into a query using typeorm I got an error of duplicate column names.

packages/data-store/src/entities/contact/PartyTypeEntity.ts
```diff
@@ -11,21 +11,21 @@ export class PartyTypeEntity {
   @PrimaryGeneratedColumn('uuid')
   id!: string
 
-  @Column('simple-enum', { name: 'type', enum: PartyTypeType, nullable: false, unique: false })
+  @Column('simple-enum', { name: 'type', enum: PartyTypeType, nullable: false })
   type!: PartyTypeType
 
-  @Column('simple-enum', { name: 'origin', enum: PartyOrigin, default: 'EXTERNAL', nullable: false, unique: false })
+  @Column('simple-enum', { name: 'party_origin', enum: PartyOrigin, default: 'EXTERNAL', nullable: false, unique: false })
```
sanderPostma, line 17: I have to differentiate between part origin and identity origin because typeorm doesn't like to join tables with the same column name

packages/data-store/src/__tests__/contact.entities.test.ts
```diff
@@ -544,32 +551,11 @@ describe('Database entities tests', (): void => {
     await expect(dbConnection.getRepository(IdentityEntity).save(identityEntity)).rejects.toThrowError('Blank metadata labels are not allowed')
   })
 
-  it('should throw error when saving identity with blank metadata value', async (): Promise<void> => {
```
sanderPostma, line 547: Redundant because saving an empty metadata fields no longer throws an error. 
As discussed on Slack, DB insertion is skipped when the metadata value is empty.

packages/data-store/src/entities/contact/ContactMetadataItemEntity.ts
```diff
@@ -0,0 +1,50 @@
...
+export class ContactMetadataItemEntity extends BaseEntity {
+  @PrimaryGeneratedColumn('uuid')
+  id!: string
+
+  @Column({ name: 'label', length: 255, nullable: false })
+  @IsNotEmpty({ message: 'Blank metadata labels are not allowed' })
+  label!: string
+
+  @Column({ name: 'valueType', nullable: false })
+  @IsNotEmpty({ message: 'valueType must not be empty' })
```
BtencateSphereon, line 17: should we add a test for this, like we do for the other checks?

BtencateSphereon: same for the identity metadata

packages/data-store/src/entities/contact/ContactMetadataItemEntity.ts
```diff
@@ -0,0 +1,50 @@
...
+  valueType!: string
+
+  @Column({ name: 'stringValue', length: 255, nullable: true })
+  stringValue?: string
+
+  @Column({ name: 'numberValue', nullable: false })
```
BtencateSphereon, line 23: it says nullable false, but the field is optional? same for all other fields and identity

packages/data-store/src/migrations/generic/8-CreateContacts.ts
```diff
@@ -0,0 +1,66 @@
+import { DatabaseType, MigrationInterface, QueryRunner } from 'typeorm'
```
BtencateSphereon, line 1: did niels not mention we should not create new migration files?

"But before we do that I would like someone to go over all existing migrations and code as we are combining different naming strategies for tables etc"

packages/data-store/src/index.ts
```diff
@@ -66,6 +67,7 @@ export const DataStoreContactEntities = [
   NaturalPersonEntity,
   ElectronicAddressEntity,
   PhysicalAddressEntity,
+  ContactMetadataItemEntity,
```
BtencateSphereon, line 70: needs to be exported further down as well

packages/data-store/src/utils/contact/MappingUtils.ts
```diff
@@ -103,7 +110,7 @@ export const contactFrom = (contact: BaseContactEntity): Contact => {
 }
 
 export const isNaturalPerson = (contact: NonPersistedContact | BaseContactEntity): contact is NonPersistedNaturalPerson | NaturalPersonEntity =>
-  'firstName' in contact && 'lastName' in contact
+  'firstName' in contact && 'lastName' in contact && !('grade' in contact) && !('dateOfBirth' in contact)
```
BtencateSphereon, line 113: is this correct? this is old code right ? there is no grade and dateOfBirth

packages/data-store/src/utils/contact/MappingUtils.ts
```diff
@@ -202,27 +224,38 @@ export const physicalAddressFrom = (physicalAddress: PhysicalAddressEntity): Phy
...
+export const identityEntityFrom = (entity: NonPersistedIdentity): IdentityEntity => {
   const identityEntity: IdentityEntity = new IdentityEntity()
-  identityEntity.alias = args.alias
-  identityEntity.roles = args.roles
-  identityEntity.identifier = correlationIdentifierEntityFrom(args.identifier)
-  identityEntity.connection = args.connection ? connectionEntityFrom(args.connection) : undefined
-  identityEntity.metadata = args.metadata ? args.metadata.map((item: NonPersistedMetadataItem) => metadataItemEntityFrom(item)) : []
-
+  identityEntity.alias = entity.alias
+  identityEntity.origin = entity.origin ?? IdentityOrigin.EXTERNAL
```
BtencateSphereon, line 237: just making a comment here to check the default value structure

packages/data-store/src/utils/contact/MappingUtils.ts
```diff
@@ -231,20 +264,74 @@ export const identityFrom = (identity: IdentityEntity): Identity => {
   }
 }
 
-export const metadataItemEntityFrom = (item: NonPersistedMetadataItem): IdentityMetadataItemEntity => {
-  const metadataItemEntity: IdentityMetadataItemEntity = new IdentityMetadataItemEntity()
-  metadataItemEntity.label = item.label
-  metadataItemEntity.value = item.value
+const metadataItemEntityFrom = <T extends MetadataTypes, U extends { new (): any }>(
```
BtencateSphereon, line 267: should we not export this function and make it available?

packages/oid4vci-holder/src/agent/OID4VCIHolder.ts
```diff
@@ -277,6 +278,7 @@ export class OID4VCIHolder implements IAgentPlugin {
         filter: [
           {
             identities: {
+              origin: IdentityOrigin.EXTERNAL,
```
BtencateSphereon, line 281: just making a comment for the default structure

BtencateSphereon (changes requested):

packages/contact-manager/src/agent/ContactManager.ts
```diff
@@ -176,6 +177,9 @@ export class ContactManager implements IAgentPlugin {
   /** {@inheritDoc IContactManager.cmAddIdentity} */
   private async cmAddIdentity(args: AddIdentityArgs, context: RequiredContext): Promise<Identity> {
     const { contactId, identity } = args
+    if (!identity.origin) {
```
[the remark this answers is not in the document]
sanderPostma: I removed these lines. In the entity we have
>   @Column('simple-enum', { name: 'identity_origin', enum: IdentityOrigin, **nullable: false** })
  origin!: IdentityOrigin

I did add 
> NOT NULL DEFAULT 'EXTERNAL'`
to the migrations though, otherwise we could never add this to a table with existing record.

packages/data-store/src/contact/ContactStore.ts
```diff
@@ -720,4 +711,59 @@ export class ContactStore extends AbstractContactStore {
...
+        metadataCondition.numberValue = metadata.value
+        break
+      case 'boolean':
+        metadataCondition.boolValue = metadata.value
+        break
+      case 'object':
+        if (metadata.value instanceof Date) {
+          metadataCondition.dateValue = metadata.value
+        } else {
+          throw new Error(`Unsupported object type: ${Object.prototype.toString.call(metadata.value).slice(8, -1)}`)
```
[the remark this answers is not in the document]
sanderPostma: fixed

packages/data-store/src/entities/contact/ContactMetadataItemEntity.ts
sanderPostma: Fixed
Apparently this attribute doesn't do anything atm. The migrations file did not contain these errors so, tests like "should get a party by contact metadata" didn't fail.

packages/data-store/src/entities/contact/ContactMetadataItemEntity.ts
sanderPostma: I've added extra tests in , but not for blank valueType because from the API there is no code path do achieve this.

packages/data-store/src/migrations/generic/8-CreateContacts.ts
sanderPostma: I interpreted that as a future TODO. We already have multiple migration files. I'm happy to flatten them, but "someone has to go over all existing migrations and code" when we have time.
For now Maas created 8-CreateContacts.ts, I created 9-CreateContacts.ts
Separate files does have the advantage of better separation of concern => better readability rather than one giant worm of SQL lines.

packages/data-store/src/utils/contact/MappingUtils.ts
sanderPostma: Orphaned code, good catch. 
Removed.

packages/data-store/src/utils/contact/MappingUtils.ts
sanderPostma: "?? IdentityOrigin.EXTERNAL" was in the branch when I took it over, but since it's mandatory in the entity, I've also made it mandatory in the DTO. It will break compatibility, but this is the way to bring to to the developer's attention.
This is the way...

packages/data-store/src/utils/contact/MappingUtils.ts
sanderPostma: It's not really meant to. There are the wrappers identityMetadataItemEntityFrom and contactMetadataItemEntityFrom wrappers to make the API more clear

packages/contact-manager/src/agent/ContactManager.ts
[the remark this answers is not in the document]
BtencateSphereon: you can because you can write a query that adds a value to existing ones

packages/data-store/src/entities/contact/ContactMetadataItemEntity.ts
BtencateSphereon: you should be able too? we are doing that for all of these blank values? the code should check for blank values optherwise the user justs sees a sql error?

packages/data-store/src/migrations/generic/8-CreateContacts.ts
BtencateSphereon: please check with niels

packages/contact-manager/src/agent/ContactManager.ts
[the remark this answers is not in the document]
sanderPostma: Removed DEFAULT 'EXTERNAL it for PartyType because there is only a CREATE TABLE for that
For identity I changed to to 
```
    await queryRunner.query(`ALTER TABLE "Identity" ADD COLUMN "identity_origin" "public"."IdentityOrigin_type_enum" DEFAULT 'EXTERNAL' NOT NULL`)
    await queryRunner.query(`ALTER TABLE "Identity" ALTER COLUMN "identity_origin" DROP DEFAULT`)
```

packages/data-store/src/index.ts
sanderPostma: Fixed

packages/oid4vci-holder/src/agent/OID4VCIHolder.ts
sanderPostma: Commented it out with TODO pending feedback

packages/data-store/src/entities/contact/ContactMetadataItemEntity.ts
sanderPostma: You cannot do that, valueType is not present in the DTO and not reachable from the contactMetadataItemEntityFrom(metadataItem)

packages/data-store/src/utils/contact/MappingUtils.ts
BtencateSphereon: ok yes i meant the partyType. if i look at the code, we use different approaches to set default values for the partyType and the identity. and i wonder why

packages/data-store/src/contact/ContactStore.ts
```diff
@@ -720,4 +711,59 @@ export class ContactStore extends AbstractContactStore {
...
+        conditionObject[key] = this.processCondition(value)
+      } else {
+        conditionObject[key] = value
+      }
+    })
+
+    return conditionObject
+  }
+
+  private buildMetadataCondition(metadata: PartialMetadataItem<MetadataTypes>): FindOptionsWhere<IdentityMetadataItemEntity> {
```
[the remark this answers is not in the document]
sanderPostma: Created generic version

packages/data-store/src/migrations/generic/8-CreateContacts.ts
sanderPostma: See Jira remark

packages/contact-manager/src/agent/ContactManager.ts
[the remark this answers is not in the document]
BtencateSphereon: PartyType still has a default in the entity. While the migrations do not have a default on PartyType. so in the entity it should be removed.

This is reversed for the indentity_origin. the entity does not have one but the migrations have one.

BtencateSphereon:
tests are failing

packages/contact-manager/src/agent/ContactManager.ts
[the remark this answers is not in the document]
sanderPostma: > PartyType still has a default in the entity
Fixed

packages/data-store/src/contact/ContactStore.ts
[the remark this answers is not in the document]
sanderPostma: Added comment // slice to extract type from string [object String]

packages/contact-manager/src/agent/ContactManager.ts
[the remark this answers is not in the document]
sanderPostma: Next try...

BtencateSphereon (approved):

PR #175, opened 2024-03-20
feature/CWALL-155

maikel-maas:
Migration files #8 are in CWALL-153


packages/data-store/src/migrations/postgres/1690925872592-CreateContacts.ts
```diff
@@ -85,6 +85,12 @@ export class CreateContacts1690925872592 implements MigrationInterface {
       `INSERT INTO "BaseContact"(id, legal_name, display_name, party_id, created_at, last_updated_at, type) SELECT id, name, alias, id, created_at, last_updated_at, 'Organization' FROM "Contact"`,
     )
     await queryRunner.query(`DROP TABLE "Contact"`)
+    await queryRunner.query(
+        `CREATE TYPE "public"."partyOrigin_type_enum" AS ENUM('internal', 'external')`
+    )
+    await queryRunner.query(
+        `ALTER TABLE "PartyType" ADD COLUMN "origin" "public"."partyOrigin_type_enum" NOT NULL`
```
nklomp, line 92: Why are you altering a table, that is literally created in this migration file?

packages/data-store/src/entities/contact/PartyTypeEntity.ts
```diff
@@ -14,6 +24,9 @@ export class PartyTypeEntity {
   @Column('simple-enum', { name: 'type', enum: PartyTypeEnum, nullable: false, unique: false })
   type!: PartyTypeEnum
 
+  @Column('simple-enum', { name: 'origin', enum: PartyOrigin, nullable: true, unique: false })
```
nklomp, line 27: I think we discussed in the sprint start to have a default (external) and make it non-nullable

packages/data-store/src/migrations/sqlite/1690925872693-CreateContacts.ts
```diff
@@ -149,6 +149,9 @@ export class CreateContacts1690925872693 implements MigrationInterface {
       `INSERT INTO "BaseContact"(id, legal_name, display_name, party_id, created_at, last_updated_at, type) SELECT id, name, alias, id, created_at, last_updated_at, 'Organization' FROM "Contact"`,
     )
     await queryRunner.query(`DROP TABLE "Contact"`)
+    await queryRunner.query(
+        `ALTER TABLE "PartyType" ADD COLUMN "origin" varchar CHECK( "origin" IN ('internal', 'external') )`
```
nklomp, line 153: Why are you altering a table, that is literally created in this migration file?

packages/data-store/src/types/contact/contact.ts
```diff
@@ -150,9 +150,15 @@ export type Contact = NaturalPerson | Organization
 export type NonPersistedContact = NonPersistedNaturalPerson | NonPersistedOrganization
 export type PartialContact = PartialNaturalPerson | PartialOrganization
 
+export enum PartyOrigin {
+  INTERNAL = 'internal',
```
nklomp, line 154: Given it is likely this enum will be used in DTOs have the casing the same!

packages/data-store/src/types/contact/contact.ts
```diff
@@ -150,9 +150,15 @@ export type Contact = NaturalPerson | Organization
...
+  EXTERNAL = 'external',
+}
+
 export type PartyType = {
   id: string
   type: PartyTypeEnum
+  origin?: PartyOrigin
```
nklomp, line 161: We discussed in the sprint start to make them mandatory with a default of external

nklomp (changes requested):
See remarks about what was discussed during sprint start as well as altering tables that are created in the same file

packages/data-store/src/migrations/postgres/1690925872592-CreateContacts.ts
maikel-maas: Adding the column at table creation

packages/data-store/src/migrations/sqlite/1690925872693-CreateContacts.ts
maikel-maas: Adding the column at table creation

packages/data-store/src/entities/contact/PartyTypeEntity.ts
maikel-maas: Made it non-nullable with external as default

packages/data-store/src/types/contact/contact.ts
maikel-maas: Made all lower case

packages/data-store/src/types/contact/contact.ts
maikel-maas: Made origin mandatory in PartyType

nklomp (approved):
Making the enums all caps would have made more sense to me, but approving

PR #183, opened 2024-04-10
feat: add kb-jwt to sd-jwt

TimoGlastra:
This adds an optional `kbJwt` to a decoded SD JWT. I found an issue in SIOP-OID4VP if you are not using ID Token and only VP, it can't extract the nonce from an SD-JWt as it tries to extract it from the signed SD-JWT, but it should take it from the KB-JWT.

To avoid scattered decoding logic, adding it here seems the most logical. Will follow up with a PR to SIOP-OID4VP


nklomp (approved):
LGTM

PR #184, opened 2024-04-23
feature/SDK-10

maikel-maas:


packages/ebsi-authorization-client/src/agent/EBSIAuthorizationClient.ts
```diff
@@ -0,0 +1,363 @@
...
+        privateKeyHex: keyPair.privateKeyHex,
+        payloadVc: credential,
+      },
+      context,
+    )
+
+    const vpJwt = await createVPJwt(
+      {
+        keyPair,
+        did: this.configuration.issuer.toString(),
```
nklomp, line 66: This should be an argument to the method, not something configured

packages/ebsi-authorization-client/src/agent/EBSIAuthorizationClient.ts
```diff
@@ -0,0 +1,363 @@
...
+  readonly methods: IEBSIAuthorizationClient = {
+    getOIDProviderMetadata: this.getOIDProviderMetadata.bind(this),
+    getOIDProviderJwks: this.getOIDProviderJwks.bind(this),
+    getPresentationDefinition: this.getPresentationDefinition.bind(this),
+    getAccessToken: this.getAccessToken.bind(this),
+    initiateSIOPDidAuthRequest: this.initiateSIOPDidAuthRequest.bind(this),
+    createSIOPSession: this.createSIOPSession.bind(this),
+    createOAuth2Session: this.createOAuth2Session.bind(this),
+  }
+  private configuration: GetOIDProviderMetadataSuccessResponse
```
nklomp, line 37: This is a very confusing name for something which is of type SuccessResponse and apparently OID metadata

packages/ebsi-authorization-client/src/agent/EBSIAuthorizationClient.ts
```diff
@@ -0,0 +1,363 @@
...
+
+  //TODO Pass the configuration manually or call init
+  constructor(readonly baseUrl: string = 'https://api-pilot.ebsi.eu/authorisation/v4/') {
+    // TODO call configuration endpoint
```
nklomp, line 41: Not from constructor. Rather would see a lazy method being called as a first action of any method that needs the metdata

packages/ebsi-authorization-client/src/agent/EBSIAuthorizationClient.ts
```diff
@@ -0,0 +1,363 @@
...
+      },
+      context,
+    )
+
+    let descriptorMap: Descriptor[] = []
+
+    if (definitionId === ScopeByDefinition.didr_invite_presentation) {
+      descriptorMap = [
+        {
+          id: 'didr_invite_credential',
```
nklomp, line 79: The only difference in these descriptors is the id in 2 places. Suggest to create a function and pass in the id argument

packages/ebsi-authorization-client/src/agent/EBSIAuthorizationClient.ts
```diff
@@ -0,0 +1,363 @@
...
+    })
+    if ('status' in tokenResponse) {
+      throw new Error(JSON.stringify(tokenResponse))
+    }
+    return tokenResponse
+  }
+
+  private async getOIDProviderMetadata(): Promise<GetOIDProviderMetadataResponse> {
+    return await (
+      await fetch(`${this.baseUrl}/.well-known/openid-configuration`, {
```
nklomp, line 137: We should have generic fetch support, which does proper error handling. Either in ssi-types, sdk-core or express-support from the top of my head

packages/ebsi-authorization-client/src/agent/EBSIAuthorizationClient.ts
```diff
@@ -0,0 +1,363 @@
...
+    throw new Error('Verifiable Credential not defined')
+  }
+
+  const privateKeyJWT = await exportJWK(u8a.fromString(keyPair.privateKeyHex, 'hex'))
+  const publicKeyJWT = await exportJWK(u8a.fromString(keyPair.publicKeyHex, 'hex'))
+
+  let verifiableCredential: string[]
+
+  if (vc === 'empty') {
+    verifiableCredential = []
```
nklomp, line 259: How would no VCs result in a valid VP where the definitions already made clear it expects 1 VC?

packages/ebsi-authorization-client/src/agent/EBSIAuthorizationClient.ts
```diff
@@ -0,0 +1,363 @@
...
+
+  if (!keyPair || !keyPair.privateKeyHex || !keyPair.publicKeyHex) {
+    throw new Error(`Private key is required`)
+  }
+
+  if (!args.vc) {
+    throw new Error('Verifiable Credential not defined')
+  }
+
+  const privateKeyJWT = await exportJWK(u8a.fromString(keyPair.privateKeyHex, 'hex'))
```
nklomp, line 253: Never ever handle private keys outside the KMS

packages/ebsi-authorization-client/src/agent/EBSIAuthorizationClient.ts
```diff
@@ -0,0 +1,363 @@
...
+    await signingKey.export({
+      type: 'JsonWebKey2020',
+      privateKey: true,
+    }),
+    {
+      detached: false,
+      header: jwtHeader,
+    },
+  )
+  const vpSigner = vpKey.signer()
```
nklomp, line 319: Don;t use external libs for this. The agent is perfectly capable of creating VCs and VPs of course!

packages/ebsi-authorization-client/src/types/IEBSIAuthorizationClient.ts
```diff
@@ -0,0 +1,310 @@
+import { IAgentContext, ICredentialPlugin, IDIDManager, IKeyManager, IPluginMethodMap } from '@veramo/core'
+import { Format, PresentationDefinitionV2, PresentationSubmission } from '@sphereon/pex-models'
+
+/**
+ * The OpenID scope
+ * @readonly
+ * @enum {string}
+ */
+export enum OpenIDScope {
+  didr_write = 'openid%20didr_write',
```
nklomp, line 10: Hmzz, I don;t like that. The openId scope is litterally only "openid". Then EBSI is defining other scopes as well. Their eindpoints expect 2 scopes. Calling these OpenID scopes and then already combining them is not nice. Simply ensure the scope is a seperate enum and ensure the openid scope is always present when constructing the scope

packages/ebsi-authorization-client/src/types/IEBSIAuthorizationClient.ts
```diff
@@ -0,0 +1,310 @@
...
+ * @property {string[]} [request_parameter_supported] Boolean value specifying whether the OP supports use of the request parameter, with true indicating support
+ * @property {string[]} [token_endpoint_auth_methods_supported] JSON array containing a list of client authentication methods supported by this token endpoint
+ * @property {{ authorization_endpoint: string[] }} [request_authentication_methods_supported] A JSON Object defining the client authentications supported for each endpoint
+ * @property {string[]} [vp_formats_supported] An object containing a list of key value pairs, where the key is a string identifying a credential format supported by the AS
+ * @property {(URL[] | string[])} [subject_syntax_types_supported]  A JSON array of strings representing URI scheme identifiers and optionally method names of supported Subject Syntax Types
+ * @property {string[]} [subject_trust_frameworks_supported] A JSON array of supported trust frameworks.
+ * @property {string[]} [id_token_types_supported] A JSON array of strings containing the list of ID Token types supported by the OP
+ */
+export interface GetOIDProviderMetadataSuccessResponse {
```
nklomp, line 68: Why are we redefining what we have available in our SIOP libs?

packages/ebsi-authorization-client/src/types/IEBSIAuthorizationClient.ts
```diff
@@ -0,0 +1,310 @@
...
+ * @typedef Key
+ * @type {object}
+ * @property {string} [kty]
+ * @property {string} [crv]
+ * @property {string} [alg]
+ * @property {string} [x]
+ * @property {string} [y]
+ * @property {string} [kid]
+ */
+export interface Key {
```
nklomp, line 100: Why do we need to define this. Not like we are not using keys to elsewhere to begin with. Next to that this is actually a JWK it seems, so please call it a JWK then

packages/ebsi-authorization-client/src/types/IEBSIAuthorizationClient.ts
```diff
@@ -0,0 +1,310 @@
...
+ * @type {object}
+ * @property {string} [ake1_enc_payload] Encrypted payload with user's public key
+ * @property {string} [ake1_nonce] Nonce used during the authentication process
+ * @property {string} [did] API DID
+ * @property {string} [kid] Trusted App KID
+ * @property {number} [iat] Issued at
+ * @property {number} [exp] Expires at
+ * @property {string} [iss] Issuer (Authorisation API)
+ */
+export interface Ake1SigPayload {
```
nklomp, line 231: What is this doing here?

nklomp (changes requested):
This will need a significant amount of changes. Never ever require private keys in methods, except for the KMS no agent plugin can access private keys

packages/ebsi-authorization-client/src/agent/EBSIAuthorizationClient.ts
maikel-maas: Extracted into a method: getDescriptorMap(...)

packages/ebsi-authorization-client/src/types/IEBSIAuthorizationClient.ts
maikel-maas: Changed the name of the enum to EBSIScope and added openid scope directly to the request

packages/ebsi-authorization-client/src/agent/EBSIAuthorizationClient.ts
maikel-maas: Searching for the key in kms using a kid now

packages/ebsi-authorization-client/src/agent/EBSIAuthorizationClient.ts
maikel-maas: Something inherited from the EBSI code base. Removed

packages/ebsi-authorization-client/src/agent/EBSIAuthorizationClient.ts
maikel-maas: Removed third-party libraries and signing JWTs with the agent now

packages/ebsi-authorization-client/src/agent/EBSIAuthorizationClient.ts
maikel-maas: Passing and argument to the method now

packages/ebsi-authorization-client/src/agent/EBSIAuthorizationClient.ts
maikel-maas: Removed based on the comment below

packages/ebsi-authorization-client/src/agent/EBSIAuthorizationClient.ts
```diff
@@ -0,0 +1,363 @@
...
+  GetPresentationDefinitionArgs,
+  GetPresentationDefinitionResponse,
+  IEBSIAuthorizationClient,
+  InitiateSIOPDidAuthRequestArgs,
+  InitiateSIOPDidAuthRequestResponse,
+} from '../types/IEBSIAuthorizationClient'
+import { uuid } from 'uuidv4'
+import { Descriptor } from '@sphereon/pex-models'
+import { JWTHeader, JWTPayload } from 'did-jwt'
+import { exportJWK, importJWK, JWK, SignJWT } from 'jose'
```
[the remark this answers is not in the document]
maikel-maas: Removed jose

packages/ebsi-authorization-client/src/agent/EBSIAuthorizationClient.ts
```diff
@@ -0,0 +1,363 @@
...
+import * as u8a from 'uint8arrays'
+import { JsonWebKey, JsonWebKey2020 } from '@transmute/json-web-signature'
```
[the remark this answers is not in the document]
maikel-maas: Removed @transmute/json-web-signature

packages/ebsi-authorization-client/src/types/IEBSIAuthorizationClient.ts
maikel-maas: Importing JWK from @sphereon/did-auth-siop

packages/ebsi-authorization-client/src/agent/EBSIAuthorizationClient.ts
maikel-maas: There only one of the 3 libraries that has cross-fetch as a dependency: ssi-sdk-core. It's used specifically to download images. What do I miss?

packages/ebsi-authorization-client/src/types/IEBSIAuthorizationClient.ts
maikel-maas: This is the response from the EBSI RP

packages/ebsi-authorization-client/src/agent/EBSIAuthorizationClient.ts
```diff
@@ -0,0 +1,387 @@
...
+  IEBSIAuthorizationClient,
+  InitiateSIOPDidAuthRequestArgs,
+  InitiateSIOPDidAuthRequestResponse,
+} from '../types/IEBSIAuthorizationClient'
+import { uuid } from 'uuidv4'
+
+export class EBSIAuthorizationClient implements IAgentPlugin {
+  readonly schema = schema.IEBSIAuthorizationClient
+  readonly methods: IEBSIAuthorizationClient = {
+    getOIDProviderMetadata: this.getOIDProviderMetadata.bind(this),
```
nklomp, line 24: Please prefix them all with 'ebsi', like we do in other agent methods. Remember that these methods will all become available if you include this plugin in an agent. Then a method called "getAccessToken" has no relationship for a developer anymore to EBSI. Hence why we prefix plugin methods, so they make sense to an external developer and are also grouped together

packages/ebsi-authorization-client/src/agent/EBSIAuthorizationClient.ts
```diff
@@ -0,0 +1,387 @@
...
+    return tokenResponse
+  }
+
+  async siop(args: { scope: EBSIScope; callbackUrl: string; did: string }, context: IRequiredContext) {
+    const { scope, callbackUrl, did } = args
+
+    const authRequest = await this.initiateSIOPDidAuthRequest({ scope })
+
+    // TODO add proper error handling
+    if (!(typeof authRequest === 'string')) {
```
nklomp, line 93: Why not !==?

packages/ebsi-authorization-client/src/agent/EBSIAuthorizationClient.ts
maikel-maas: Prefixed the methods with ebsiAuth

packages/ebsi-authorization-client/src/agent/EBSIAuthorizationClient.ts
```diff
@@ -0,0 +1,418 @@
...
+      id: uuid(),
+      definition_id: id,
+      descriptor_map: descriptorMap,
+    }
+
+    const tokenResponse = await this.getAccessToken({
+      grant_type: 'vp_token',
+      vp_token: vpJwt,
+      presentation_submission: presentationSubmission,
+      scope: EBSIScope.didr_invite,
```
jcmelati, line 207: Should this `scope` come from args instead of being hardcoded?

maikel-maas: Retrieving the values from args now

packages/ebsi-authorization-client/src/types/IEBSIAuthorizationClient.ts
maikel-maas: Not needed anymore

PR #185, opened 2024-04-24
feature/CWALL-166

BtencateSphereon:


packages/oid4vci-holder/src/agent/OID4VCIHolderService.ts
```diff
@@ -204,28 +232,329 @@ export const verifyCredential = async (args: IVerifyCredentialArgs, context: Req
...
+    case 'Secp256k1':
+      return SignatureAlgorithmEnum.ES256K
+    default:
+      throw new Error(`Key type '${type}' not supported`)
+  }
+}
+
+// TODO improve this conversion for jwt and jsonld, not a fan of current structure
+export const keyTypeFromCryptographicSuite = (args: KeyTypeFromCryptographicSuiteArgs): TKeyType => {
+  // TODO async?
```
nklomp, line 543: Async for a mapping function?

nklomp (approved):
Some of the moved code for getting identifiers and preferences will need to be refactored, but overall looks good, as that code is mostly just moved from wallet to SDK

packages/oid4vci-holder/src/agent/OID4VCIHolderService.ts
BtencateSphereon: no this is an old comment, should have been removed

PR #191, opened 2024-04-26
feature/SDK-11 fixing inconsistency between UniRegistrar and ssi-sdk deactivation did function

sksadjad:
added:
- deactivateDidEndpoint function according to decentralized)-identity's universal-registrar
changed:
- renamed previous method as deleteDidEndpoint and marked it as deprecated

### DIFF
--- a/packages/uni-resolver-registrar-api/src/api-functions.ts
+++ b/packages/uni-resolver-registrar-api/src/api-functions.ts
@@ -9,6 +9,7 @@ import { v4 } from 'uuid'
 import {
   CreateState,
   DidRegistrationCreateRequest,
+  DidRegistrationDeactivateRequest,
   DidStateValue,
   ICreateDidEndpointOpts,
   IGlobalDidWebEndpointOpts,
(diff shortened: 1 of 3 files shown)


packages/uni-resolver-registrar-api/src/api-functions.ts
```diff
@@ -167,7 +168,13 @@ export function resolveDidEndpoint(router: Router, context: IRequiredContext, op
...
 }
 
-export function deactivateDidEndpoint(router: Router, context: IRequiredContext, opts?: ISingleEndpointOpts) {
+/**
+ * @deprecated
+ * @param router
+ * @param context
+ * @param opts
+ */
+export function deleteDidEndpoint(router: Router, context: IRequiredContext, opts?: ISingleEndpointOpts) {
```
jcmelati, line 177: Should this endpoint be removed? It seems that we ended up with duplicated code as `deleteDidEndpoint` and `deactivateDidEndpoint` receive the same parameters and call the same function.

nklomp: No, we should create a ticket to make it configurable. I want to also have endpoints that conform to proper REST standards, as the DIF/W3C VC APIs typically are terrible in design

nklomp: Also we have actual customers using it (for instance GS1)

packages/uni-resolver-registrar-api/src/api-functions.ts
```diff
@@ -167,7 +168,13 @@ export function resolveDidEndpoint(router: Router, context: IRequiredContext, op
   })
 }
 
-export function deactivateDidEndpoint(router: Router, context: IRequiredContext, opts?: ISingleEndpointOpts) {
+/**
+ * @deprecated
```
nklomp, line 172:
```suggestion
```

nklomp (approved):

PR #192, opened 2024-05-01
feature/CWALL-208

BtencateSphereon:


maikel-maas (approved):
LGTM

packages/data-store/src/types/contact/contact.ts
```diff
@@ -214,24 +209,29 @@ export type ElectronicAddressType = 'email' | 'phone'
...
 }
 
-export enum PartyTypeEnum {
+export enum PartyTypeType {
   NATURAL_PERSON = 'naturalPerson',
   ORGANIZATION = 'organization',
 }
+
+export enum PartyOrigin {
+  INTERNAL = 'internal',
```
nklomp, line 235: Please make enum names and values the same in case. We have problems with these when they are used in DTOs/REST APIs

nklomp (changes requested):
enum name and value case are different

packages/data-store/src/types/contact/contact.ts
BtencateSphereon: as discussed, i made this one the same, name and value are uppercase now

nklomp (approved):

nklomp:
LGTM

PR #198, opened 2024-06-03
bugfix/CWALL-229

maikel-maas:


BtencateSphereon (changes requested):
removing the constraints make 0 sense to me. Can you explain why it is needed to remove these

PR #201, opened 2024-06-05
Chore/cwall 200 based on cwall 175

sksadjad:
changed the name of the exposed functions of issuerBranding module


nklomp (approved):
Approve. Please next time make sure to adjust the base beforehand, so I don't have to. It looked like a massive PR, whilst in the end it is 2 files

nklomp (approved):

PR #204, opened 2024-06-07
feature/SPRIND-3

BtencateSphereon:


packages/sd-jwt/src/action-handler.ts
```diff
@@ -0,0 +1,227 @@
...
+ */
+export class SDJwtPlugin implements IAgentPlugin {
+  readonly schema = schema.ISDJwtPlugin
+
+  constructor(private algorithms: SdJWTImplementation) {}
+
+  // map the methods your plugin is declaring to their implementation
+  readonly methods: ISDJwtPlugin = {
+    createSdJwtVc: this.createSdJwtVc.bind(this),
+    createSdJwtVcPresentation: this.createSdJwtVcPresentation.bind(this),
```
nklomp, line 35: I know this is from the original plugin, but suggest to remove Vc from the Presentation function names. It is unneeded and confusing

packages/sd-jwt/src/action-handler.ts
```diff
@@ -0,0 +1,227 @@
...
+   * @returns the key to sign the SD-JWT
+   */
+  private async getSignKey(issuer: string, context: IRequiredContext) {
+    const identifier = await context.agent.didManagerGet({
+      did: issuer.split('#')[0],
+    })
+    const doc = await mapIdentifierKeysToDocWithJwkSupport(identifier, 'assertionMethod', context)
+    if (!doc || doc.length === 0) throw new Error('No key found for signing')
+    const key = doc.find((key) => key.meta.verificationMethod.id === issuer)
+    if (!key) throw new Error('No key found with the given id')
```
nklomp, line 85: Adjust to our use of accolades and include the issuer value in the error

packages/sd-jwt/src/action-handler.ts
```diff
@@ -0,0 +1,227 @@
...
+   * @returns A signed SD-JWT presentation.
+   */
+  async createSdJwtVcPresentation(args: ICreateSdJwtVcPresentationArgs, context: IRequiredContext): Promise<ICreateSdJwtVcPresentationResult> {
+    const cred = await SDJwt.fromEncode(args.presentation, this.algorithms.hasher)
+    const claims = await cred.getClaims<Claims>(this.algorithms.hasher)
+    let holderDID: string
+    // we primarly look for a cnf field, if it's not there we look for a sub field. If this is also not given, we throw an error since we can not sign it.
+    if (claims.cnf?.jwk) {
+      const key = claims.cnf.jwk
+      holderDID = `did:jwk:${encodeJoseBlob(key)}#0`
```
nklomp, line 117: Hmzz, this seems a bit naive. We probably should place a TODO saying that we should convert the JWK to hex and search for the appropriate key and associated DID. Certainly doesn't apply to did:jwk only, as you can represent any DID key as a JWK. So whenever you encounter a JWK it doesn't mean it had to come from a did:jwk in the system. It just can always be represented as a did:jwk

packages/sd-jwt/src/action-handler.ts
```diff
@@ -0,0 +1,227 @@
...
+    const decodedVC = await sdjwt.decode(`${data}.${signature}`)
+    const issuer: string = ((decodedVC.jwt as Jwt).payload as Record<string, unknown>).iss as string
+    if (!issuer.startsWith('did:')) {
+      throw new Error('invalid_issuer: issuer must be a did')
+    }
+    const didDoc = await context.agent.resolveDid({ didUrl: issuer })
+    if (!didDoc) {
+      throw new Error('invalid_issuer: issuer did not resolve to a did document')
+    }
+    const didDocumentKey = didDoc.didDocument?.verificationMethod?.find((key) => key.id)
```
nklomp, line 196: This basically does nothing except for checking there is a verificationMethod with an id. That is way too naive. It should be checking for an assertionMethod. Suggest to create a TODO we need to fix this

packages/sd-jwt/src/action-handler.ts
```diff
@@ -0,0 +1,227 @@
...
+    if (!didDocumentKey) {
+      throw new Error('invalid_issuer: issuer did document does not include referenced key')
+    }
+    //TODO: in case it's another did method, the value of the key can be also encoded as a base64url
+    const key = didDocumentKey.publicKeyJwk as JsonWebKey
```
nklomp, line 201: This needs more checking. Because the above TODO really should be a fixme. This for sure will give issues for some DID methods as they do not expose the keys as publicKeyJwk

nklomp (changes requested):
Although of course not our code, made some initial comments. Most of it doesn't have to be addressed right now. More comments about creating some tickets, as some code fragments for sure will go and bite us soon

packages/sd-jwt/src/action-handler.ts
BtencateSphereon: renamed

packages/sd-jwt/src/action-handler.ts
BtencateSphereon: fixed and added

packages/sd-jwt/src/action-handler.ts
BtencateSphereon: added a TODO

packages/sd-jwt/src/action-handler.ts
BtencateSphereon: added a TODO

packages/sd-jwt/src/action-handler.ts
BtencateSphereon: changed to a FIXME and added a comment. is this something that should be fixed now and what approach can we take to fix this?

nklomp: No we can skip it for now

nklomp (approved):

nklomp:
PR approved

PR #206, opened 2024-06-11
feature/SPRIND-4

BtencateSphereon:


packages/oid4vci-holder/src/agent/OID4VCIHolderService.ts
```diff
@@ -173,7 +176,7 @@ export const verifyCredentialToAccept = async (args: VerifyCredentialToAcceptArg
 
   const verificationResult: VerificationResult = await verifyCredential(
     {
-      credential: credential as VerifiableCredential,
+      credential: wrappedVC.original as VeramoW3CVerifiableCredential, // FIXME figure out what type needs to go where
```
nklomp, line 179: Why did you make this change? `credential` comes from the actual VCI response, which in all cases will be the actual original credential

packages/oid4vci-holder/src/agent/OID4VCIHolderService.ts
```diff
@@ -193,21 +196,28 @@ export const verifyCredentialToAccept = async (args: VerifyCredentialToAcceptArg
...
         error: 'Invalid JWT VC',
-        errorDetails: `JWT VC could was not valid with policies: ${JSON.stringify(args.policies)}`,
-      }),
-      subResults: [],
+        errorDetails: `JWT VC was not valid with policies: ${JSON.stringify(policies)}`,
+      })
     }
   } else {
-    const subResults: Array<VerificationSubResult> = []
+    //TODO look at what this is doing and make it simple and readable
```
nklomp, line 220: Please leave existing things in place. Sub results can come from for instance status list verification happening

packages/oid4vci-holder/src/types/IOID4VCIHolder.ts
```diff
@@ -333,18 +334,10 @@ export enum SupportedDidMethodEnum {
 
 export type VerificationResult = {
   result: boolean
-  source: WrappedVerifiableCredential | WrappedVerifiablePresentation
```
nklomp, line 336: Please leave existing things in place. They really are needed

nklomp (changes requested):
Please leave existing objects in place. They are really needed and will for instance be merged when statuslists are being used

packages/oid4vci-holder/src/agent/OID4VCIHolderService.ts
BtencateSphereon: because that type is not the correct one, we are using all sorts of casts which are not correct ones

packages/oid4vci-holder/src/agent/OID4VCIHolderService.ts
BtencateSphereon: ok but they are not using nor filled ever

packages/oid4vci-holder/src/types/IOID4VCIHolder.ts
BtencateSphereon: also not ever used

packages/oid4vci-holder/src/agent/OID4VCIHolderService.ts
nklomp: Okay but the original question stands. We have something incoming which is the original thing in the original code, then changing that to something which is constructed, needs a good reason. That you might not know how/which types to use at present might not be the best argument IMO. The problem simply stems from using multiple different APIs from different parties and extending these over time. As part of the OID4VC refactoring we hopefully start next quarter these things will be addressed.

I liked the original approach with an OriginalVerifiableCredential way better then now the VeramoW3CVerifiableCredential for instance

packages/oid4vci-holder/src/agent/OID4VCIHolderService.ts
nklomp: Not in this library no. But as mentioned trustlist implementations can fill these objects. This ensures we expose the interfaces as how they can be used by other implementations

packages/oid4vci-holder/src/types/IOID4VCIHolder.ts
nklomp: Okay but then we need a fixme/todo

packages/oid4vci-holder/src/agent/OID4VCIHolderService.ts
BtencateSphereon: reverted

packages/oid4vci-holder/src/types/IOID4VCIHolder.ts
BtencateSphereon: added FIXME

packages/presentation-exchange/src/functions.ts
```diff
@@ -15,26 +27,47 @@ export async function createPEXPresentationSignCallback(
...
-  }: {
-    format?: Format | 'jwt' | 'lds' | 'EthereumEip712Signature2021'
-    presentationDefinition: IPresentationDefinition
-  }) {
-    let proofFormat: ProofFormat = 'jwt'
+  function determineProofFormat(args: {format?: Format | 'jwt' | 'lds' | 'EthereumEip712Signature2021', presentationDefinition: IPresentationDefinition }): string {
+    const {format, presentationDefinition } = args
+
+    // All format arguments are optional. So if no format has been given we go for SD-JWT
     const formatOptions = format ?? args.format ?? presentationDefinition.format
```
sksadjad, line 34: I think you can delete `args.format` here

packages/ssi-types/src/mapper/credential-mapper.ts
```diff
@@ -454,22 +451,23 @@ export class CredentialMapper {
...
+      maxTimeSkewInMS?: number,
+      hasher?: Hasher
     },
   ): IVerifiableCredential {
     if (CredentialMapper.isSdJwtDecodedCredential(verifiableCredential)) {
       throw new Error('Converting SD-JWT VC to uniform VC is not supported.')
     }
     const original =
-      typeof verifiableCredential !== 'string' && CredentialMapper.hasJWTProofType(verifiableCredential)
+      typeof verifiableCredential !== 'string' && CredentialMapper.hasJWTProofType(verifiableCredential) && !CredentialMapper.isSdJwtEncoded(verifiableCredential.proof.jwt)
```
sksadjad, line 462: not sure if `verifiableCredential.proof.jwt` should be passed here

sksadjad (changes requested):
added a few comments

packages/oid4vci-holder/src/agent/OID4VCIHolder.ts
```diff
@@ -556,7 +569,7 @@ export class OID4VCIHolder implements IAgentPlugin {
     let persist = true
     const verifiableCredential = credentialToAccept.uniformVerifiableCredential as VerifiableCredential
 
-    const notificationId = credentialToAccept.credential.credentialResponse.notification_id
+    const notificationId = credentialToAccept.credentialToAccept.credentialResponse.notification_id
```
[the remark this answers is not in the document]
nklomp: No it is not., that says nothing.

packages/ssi-types/src/mapper/credential-mapper.ts
nklomp: Yes it should

packages/presentation-exchange/src/functions.ts
```diff
@@ -46,118 +66,121 @@ export async function createPEXPresentationSignCallback(
...
-          const converted = Date.parse(presentation.issuanceDate)
-          if (!isNaN(converted)) {
-            presentation.nbf = Math.floor(converted / 1000)
+      let header
+      if (proofFormat === 'jwt') {
+        header = {
+          kid: vm.id,
+        }
+        if (presentation.verifier || !presentation.aud) {
+          // TODO check external implementations if they like a single audience instead of an array
```
nklomp, line 123: Why was this done. AFAIK the original implementation was according to spec

PR #212, opened 2024-06-17
Feature/cwall 247 vp UI

sanderPostma:


nklomp:
@sanderPostma I merged in latest develop to resolve the conflicts

packages/siopv2-oid4vp-op-auth/package.json
```diff
@@ -17,22 +17,31 @@
     "@sphereon/did-auth-siop": "0.6.4",
     "@sphereon/pex": "^3.3.3",
     "@sphereon/pex-models": "2.2.4",
+    "@sphereon/ssi-sdk-ext.did-resolver-jwk": "^0.21.0",
```
nklomp, line 20: Why does it depend on did-jwk resolution all of a sudden. That should be in dev deps

packages/siopv2-oid4vp-op-auth/src/agent/DidAuthSiopOpAuthenticator.ts
```diff
@@ -11,7 +10,36 @@ import {
...
+  CorrelationIdentifierType,
+  CredentialRole,
+  Identity,
+  IdentityOrigin,
+  NonPersistedIdentity,
+  Party,
+} from '@sphereon/ssi-sdk.data-store'
+import { siopSendAuthorizationResponse, translateCorrelationIdToName } from '../services/Siopv2MachineService'
+
+const logger = Loggers.DEFAULT.options(LOGGER_NAMESPACE, { methods: [LogMethod.CONSOLE, LogMethod.DEBUG_PKG] }).get(LOGGER_NAMESPACE)
```
nklomp, line 42: Please do not provide methods when initializing, so an app can set the defaults they want instead of the SDK/lib deciding what is needed.

packages/siopv2-oid4vp-op-auth/src/link-handler/index.ts
```diff
@@ -0,0 +1,49 @@
+import { LinkHandlerAdapter } from '@sphereon/ssi-sdk.core'
+import { IMachineStatePersistence, interpreterStartOrResume } from '@sphereon/ssi-sdk.xstate-machine-persistence'
+import { IAgentContext } from '@veramo/core'
+import { Loggers, LogMethod } from '@sphereon/ssi-types'
+import { GetMachineArgs, IDidAuthSiopOpAuthenticator, LOGGER_NAMESPACE, Siopv2MachineInterpreter, Siopv2MachineState } from '../types'
+
+const logger = Loggers.DEFAULT.options(LOGGER_NAMESPACE, { methods: [LogMethod.CONSOLE, LogMethod.DEBUG_PKG] }).get(LOGGER_NAMESPACE)
```
nklomp, line 7: Do not set options from the SDK

packages/siopv2-oid4vp-op-auth/src/localization/Localization.ts
```diff
@@ -0,0 +1,49 @@
+import i18n, { Scope, TranslateOptions } from 'i18n-js'
+import memoize from 'lodash.memoize'
+import { SupportedLanguage } from '../types'
+
+class Localization {
```
nklomp, line 5: Hmzz, probably this should move to core, or better yet an i18n plugin with slight adjustments to enable other SDK modules to define there translation locations, so that they can get an instance per module with its own i18n

packages/siopv2-oid4vp-op-auth/src/services/IdentifierService.ts
```diff
@@ -0,0 +1,77 @@
...
+        createOpts: { options: { type: keyOpts.keyType, use: KeyUse.Signature, codecName: keyOpts.codecName } },
+      },
+    }))
+  const key: _ExtendedIKey = await getAuthenticationKey({ identifier, context })
+  const kid: string = key.meta.verificationMethod.id
+
+  return { identifier, key, kid }
+}
+
+export const getAuthenticationKey = async (args: GetAuthenticationKeyArgs): Promise<_ExtendedIKey> => {
```
nklomp, line 36: A better version of this method lives in oid4vci-holder. Probably these should be moved to ssi-sdk-ext.did-utls

packages/siopv2-oid4vp-op-auth/src/services/IdentifierService.ts
```diff
@@ -0,0 +1,77 @@
...
+  const { identifier, context } = args
+  const agentContext = { ...context, agent: context.agent as TAgent<IResolver & IDIDManager> }
+
+  return (
+    (await getFirstKeyWithRelation(identifier, agentContext, 'authentication', false)) ||
+    ((await getFirstKeyWithRelation(identifier, agentContext, 'verificationMethod', true)) as _ExtendedIKey)
+  )
+}
+
+export const getOrCreatePrimaryIdentifier = async (args: GetOrCreatePrimaryIdentifierArgs): Promise<IIdentifier> => {
```
nklomp, line 46: This method also lives in oid4vci-holder. Probably these should be moved to ssi-sdk-ext.did-utls

packages/siopv2-oid4vp-op-auth/src/services/IdentifierService.ts
```diff
@@ -0,0 +1,77 @@
...
+    const createOpts = opts?.createOpts ?? {}
+    createOpts.options = { codecName: 'EBSI', type: 'Secp256r1', ...createOpts }
+    opts.createOpts = createOpts
+  }
+  const identifier: IIdentifier = !identifiers || identifiers.length == 0 ? await createIdentifier({ context, opts }) : identifiers[0]
+
+  return await context.agent.didManagerGet({ did: identifier.did })
+}
+
+export const createIdentifier = async (args: CreateIdentifierArgs): Promise<IIdentifier> => {
```
nklomp, line 64: This method lives in oid4vci-holder. Probably these should be moved to ssi-sdk-ext.did-utls

nklomp (changes requested):
See remarks. We need to move some methods from VCI and SIOP into ssi-sdk-ext.did-utils. In most cases the VCI versions have been improved, so take these as the base to move and then use them from VCI and VP

packages/siopv2-oid4vp-op-auth/src/agent/DidAuthSiopOpAuthenticator.ts
sanderPostma: This is copied from oid4vci-holder so this should be addressed here as well:
https://github.com/Sphereon-Opensource/SSI-SDK/blob/d4a4e681b34e1e5acd81b8a8574c843c3eceed70/packages/oid4vci-holder/src/agent/OID4VCIHolder.ts#L94
https://github.com/Sphereon-Opensource/SSI-SDK/blob/d4a4e681b34e1e5acd81b8a8574c843c3eceed70/packages/ssi-types/src/events/index.ts#L149

packages/oid4vci-holder/src/machine/oid4vciMachine.ts
```diff
@@ -586,6 +586,9 @@ export class OID4VCIMachine {
     }
     if (opts?.requireCustomNavigationHook !== true) {
       interpreter.onTransition((snapshot: OID4VCIMachineState): void => {
+        if (opts?.stateNavigationListener === undefined) {
```
nklomp, line 589: Why? Especially since you made the stateNavigationListener optional as well.

packages/siopv2-oid4vp-op-auth/package.json
nklomp: Ping

packages/siopv2-oid4vp-op-auth/src/agent/DidAuthSiopOpAuthenticator.ts
nklomp: Ping. Point is that we should not keep repeating if it is done somewhere else as well

nklomp (changes requested):
ping

packages/siopv2-oid4vp-op-auth/package.json
sanderPostma:
```
import { KeyUse } from '@sphereon/ssi-sdk-ext.did-resolver-jwk'
```
It was already moved to devDeps, it's just for the enum

packages/siopv2-oid4vp-op-auth/src/agent/DidAuthSiopOpAuthenticator.ts
sanderPostma: I removed the methods everywhere but in the .test.ts files last week

packages/siopv2-oid4vp-op-auth/src/link-handler/index.ts
sanderPostma: I removed the methods everywhere but in the .test.ts files last week

packages/siopv2-oid4vp-op-auth/src/services/IdentifierService.ts
sanderPostma: See PR https://github.com/Sphereon-Opensource/SSI-SDK-crypto-extensions/pull/35/files

packages/siopv2-oid4vp-op-auth/src/services/IdentifierService.ts
sanderPostma: See PR https://github.com/Sphereon-Opensource/SSI-SDK-crypto-extensions/pull/35/files

packages/oid4vci-holder/src/machine/oid4vciMachine.ts
sanderPostma: I don't remember why we did that. Removed in next commit

sanderPostma:
Hmmz I see my commits from last week were not pushed yet so I understand the ping's now...

packages/siopv2-oid4vp-op-auth/src/localization/Localization.ts
nklomp: PR merged

packages/siopv2-oid4vp-op-auth/src/services/IdentifierService.ts
nklomp: PR merged

packages/siopv2-oid4vp-op-auth/src/services/IdentifierService.ts
nklomp: PR merged

sanderPostma:
This one still needs work now. A test in contact-entities is failing now after I merged in SDK-10

nklomp:
That is fixed in latest SDK-10

PR #217, opened 2024-07-02
Feature/sprind 18 add issuer branding

sksadjad:
added saving the issuer branding to the vci steps


packages/sd-jwt/src/__tests__/sd-jwt.test.ts
```diff
@@ -30,7 +30,7 @@ async function verifySignature<T>(data: string, signature: string, key: JsonWebK
 
 type AgentType = IDIDManager & IKeyManager & IResolver & ISDJwtPlugin
 
-describe('Agent plugin', () => {
+describe.skip('Agent plugin', () => {
```
nklomp, line 33: Why is this test disabled?

sksadjad: this test was failing for me and this is temporary disabled to be able to build the artifacts that I needed for my test

nklomp: Okay, but please make remarks about these and/or discuss what the problem is so we can fix it. Because I am seeing more and more tests being disabled. If that gets past a review all we are doing is going down in test coverage

packages/data-store/src/types/contact/contact.ts
```diff
@@ -13,6 +14,8 @@ export type Party = {
   physicalAddresses: Array<PhysicalAddress>
   contact: Contact
   partyType: PartyType
+  //todo: add logic for integrating branding in here. for now we're just using the type
```
nklomp, line 17: Please clarify that message way more. I cannot determine what you actually mean with it

packages/sd-jwt/src/__tests__/sd-jwt.test.ts
sksadjad: reverted the skip. and sure.

packages/data-store/src/types/contact/contact.ts
sksadjad: modified the comment and clarified it more based on the discussion that we had

nklomp (approved):
LGTM

PR #218, opened 2024-07-03
Feature/cwall 247 vp UI

sanderPostma:
Moved some logic related to getting selectable credentials


nklomp (approved):
I don't like the name selectable credentials, but given we can quite easily change that in the next refactoring, not going to block on it

PR #222, opened 2024-07-06
Feature/cwall 247 vp UI

sanderPostma:
Additional fixes


packages/siopv2-oid4vp-op-auth/src/types/machine/index.ts
```diff
@@ -101,9 +103,10 @@ export enum Siopv2MachineGuards {
   createContactGuard = 'Siopv2CreateContactGuard',
   hasContactGuard = 'Siopv2HasContactGuard',
   hasAuthorizationRequestGuard = 'Siopv2HasAuthorizationRequestGuard',
+  canDoGetSelectableCredentialsGuard = 'Siopv2CanDoGetSelectableCredentialsGuard',
   hasSelectedRequiredCredentialsGuard = 'Siopv2HasSelectedRequiredCredentialsGuard',
   siopOnlyGuard = 'Siopv2IsSiopOnlyGuard',
-  siopWithOID4VPGuard = 'Siopv2IsSiopWithOID4VPGuard',
+  canDoSelectCredentialsGuard = 'Siopv2CanDoSelectCredentialsGuard',
```
nklomp, line 109: Why would you rename something that was more clear about the protocols in play to this?

packages/siopv2-oid4vp-op-auth/src/services/Siopv2MachineService.ts
```diff
@@ -215,7 +215,7 @@ export const translateCorrelationIdToName = async (correlationId: string, contex
     filter: [{ identities: { identifier: { correlationId } } }],
   })
   if (contacts.length === 0) {
-    return Promise.reject(Error(`Unable to find contact for correlationId ${correlationId}`))
+    return correlationId
```
nklomp, line 218: Whut?!?!

packages/siopv2-oid4vp-op-auth/src/machine/Siopv2Machine.ts
```diff
@@ -374,6 +394,7 @@ const createSiopv2Machine = (opts: CreateSiopv2MachineOpts): Siopv2StateMachine
 
 export class Siopv2Machine {
   static newInstance(opts: Siopv2MachineInstanceOpts): { interpreter: Siopv2MachineInterpreter } {
+    console.log('New Siopv2Machine instance')
```
nklomp, line 397: logger

packages/siopv2-oid4vp-op-auth/src/machine/Siopv2Machine.ts
```diff
@@ -344,10 +367,7 @@ const createSiopv2Machine = (opts: CreateSiopv2MachineOpts): Siopv2StateMachine
       [Siopv2MachineStates.handleError]: {
         id: Siopv2MachineStates.handleError,
         on: {
-          [Siopv2MachineEvents.NEXT]: {
-            target: Siopv2MachineStates.error,
-          },
-          [Siopv2MachineEvents.PREVIOUS]: {
+          '': {
```
nklomp, line 370: Why the need for the empty string to begin with? There is an always transition available

packages/siopv2-oid4vp-op-auth/src/machine/Siopv2Machine.ts
```diff
@@ -74,13 +88,17 @@ const Siopv2IsSiopOnlyGuard = (_ctx: Siopv2MachineContext, _event: Siopv2Machine
   return authorizationRequestData.presentationDefinitions === undefined
 }
 
-const Siopv2IsSiopWithOID4VPGuard = (_ctx: Siopv2MachineContext, _event: Siopv2MachineEventTypes): boolean => {
-  const { authorizationRequestData } = _ctx
+const Siopv2CanDoSelectCredentialsGuard = (_ctx: Siopv2MachineContext, _event: Siopv2MachineEventTypes): boolean => {
```
nklomp, line 91: The previous name was a better name, also be aware whenever you rename machine guards and services it impacts multiple other dependent libs/products.

packages/siopv2-oid4vp-op-auth/src/machine/Siopv2Machine.ts
```diff
@@ -38,6 +39,19 @@ const Siopv2HasAuthorizationRequestGuard = (_ctx: Siopv2MachineContext, _event:
   return authorizationRequestData !== undefined
 }
 
+const Siopv2CanDoGetSelectableCredentialsGuard = (_ctx: Siopv2MachineContext, _event: Siopv2MachineEventTypes): boolean => {
```
nklomp, line 42: This is a pretty bad name. CanDoGetSelecatable.....

packages/oid4vci-holder/src/machine/oid4vciMachine.ts
```diff
@@ -62,7 +62,7 @@ const oid4vciVerificationCodeGuard = (_ctx: OID4VCIMachineContext, _event: OID4V
 
 const oid4vciCreateContactGuard = (_ctx: OID4VCIMachineContext, _event: OID4VCIMachineEventTypes): boolean => {
   const { contactAlias, hasContactConsent } = _ctx
-  return hasContactConsent && contactAlias !== undefined && contactAlias.length > 0
+  return hasContactConsent && contactAlias !== undefined && contactAlias !== null && contactAlias.length > 0
```
nklomp, line 65:
```suggestion
  return hasContactConsent && contactAlias && contactAlias.length > 0
```

nklomp (changes requested):
See remakrs. Please do not rename into something worse and be aware of the impact that renaming services/guards cause for the wallets

packages/siopv2-oid4vp-op-auth/src/machine/Siopv2Machine.ts
sanderPostma: Reverted, but personally isSiopWithOID4VPGuard doesn't tell me anything.

packages/siopv2-oid4vp-op-auth/src/machine/Siopv2Machine.ts
sanderPostma: Okay, initially I had Siopv2HasSelectableCredentialsAndContactGuard which I reverted to now. But it's a bit long and was looking for something smaller. When I add another two to more conditions to Siopv2HasSelectableCredentialsAndContactGuard  it's going to be 
Siopv2HasSelectableCredentialsAndContactAndOtherThingAndYetAnotherThingGuard  ?

packages/siopv2-oid4vp-op-auth/src/types/machine/index.ts
sanderPostma: Reverted, see later comment

packages/siopv2-oid4vp-op-auth/src/services/Siopv2MachineService.ts
sanderPostma: Reverted. My guess is it was crashing the flow before the AddContact was working properly.
Will have to retest to see if this is bugging me.

packages/siopv2-oid4vp-op-auth/src/machine/Siopv2Machine.ts
sanderPostma: fixed

packages/siopv2-oid4vp-op-auth/src/machine/Siopv2Machine.ts
sanderPostma: For what I understand, 'always' is reevaluted / checked continuously even when not in / after a transition.
'' is evaluated only once when the state is entered.
Since I experienced duplicate triggers on the navigationStateListener this seemed the better choice.
Also I don't understand why in the previous situation we only go to error for NEXT/PREVIOUS events. When something fails in a background / unattended task to should also navigate to the error page.

packages/siopv2-oid4vp-op-auth/src/services/Siopv2MachineService.ts
nklomp: Yeah, means you should read these specs, because it should be clear

packages/siopv2-oid4vp-op-auth/src/machine/Siopv2Machine.ts
nklomp: Yeah, means you should read these specs, because it should be clear. The whole point is that it is called because a RP supports certain protocols or not, instead of the name suggesting whether the wallet could do something or not

packages/siopv2-oid4vp-op-auth/src/machine/Siopv2Machine.ts
nklomp: Well it should not be talking about the other precondintions to begin with. The guard is responsible for whether VP is used or SIOP only

nklomp (approved):

PR #225, opened 2024-07-15
feature/SDK-24

sanderPostma:
DRAFT


packages/credential-store/src/agent/CredentialStore.ts
```diff
@@ -0,0 +1,257 @@
...
+    const credentials = await this.store.getCredentials({ filter })
+    return credentials.data
+  }
+
+  /** {@inheritDoc ICRManager.crmGetUniqueCredentialByIdOrHash} */
+  private async crsGetUniqueCredentialByIdOrHash(args: GetCredentialsByIdOrHashArgs): Promise<OptionalUniqueDigitalCredential> {
+    const credentials = await this.crsGetCredentials({ filter: credentialIdOrHashFilter(args.credentialRole, args.idOrHash) })
+    if (credentials.length === 0) {
+      return undefined
+    }
```
nklomp, line 88: Probably wise to log in case length > 1.
I can imagine this would happen if the issuer, holder and rp are using the same store for the same credential, but each in their own context. Then we end up with 3 versions of the same credential in the store, but with different roles.

packages/credential-store/src/types/filters.ts
```diff
@@ -0,0 +1,83 @@
...
+]
+
+/**
+ * Creates a filter for verifiable credentials with a specific role.
+ *
+ * @param credentialRole - The role to filter by (e.g., ISSUER, HOLDER).
+ * @param withFilter - Optional additional filter criteria.
+ * @returns A FindDigitalCredentialArgs array for filtering verifiable credentials by role.
+ */
+export const verifiableCredentialForRoleFilter = (
```
nklomp, line 28: Why is this only available for VCs? We need this for C, VP, P as well, as all of these can have roles

packages/oid4vci-holder/src/agent/OID4VCIHolder.ts
```diff
@@ -871,8 +871,16 @@ export class OID4VCIHolder implements IAgentPlugin {
...
     } else {
       logger.log(`Persisting credential`, persistCredential)
-      // @ts-ignore
-      const vcHash = await context.agent.dataStoreSaveVerifiableCredential({ verifiableCredential: persistCredential })
+      const vcHash = await context.agent.crsAddCredential({
+        credential: {
+          rawDocument: JSON.stringify(persistCredential),
+          credentialRole: CredentialRole.HOLDER,
+          issuerCorrelationType: CredentialCorrelationType.DID,
+          issuerCorrelationId: (persistCredential as any).issuer, // FIXME BEFORE PR
```
nklomp, line 879: FIXMEs mentioned in the test

packages/credential-store/src/agent/CredentialStore.ts
```diff
@@ -0,0 +1,257 @@
...
+
+  /**
+   * Returns a list of UniqueDigitalCredentials that match the given filter based on the claims they contain.
+   * @param args
+   */
+  private async crsGetCredentialsByClaims(args: GetCredentialsByClaimsArgs): Promise<Array<UniqueDigitalCredential>> {
+    const digitalCredentials = await this.crsGetUniqueCredentials({
+      filter: [
+        {
+          documentType: DocumentType.VC, // TODO does crmGetCredentialsByClaims need to support VPs as well?
```
nklomp: At least C should be supported, but a ticket should be created to indeed also support (V)P

packages/credential-store/src/types/ICredentialStore.ts
```diff
@@ -0,0 +1,121 @@
...
+
+export type AddCredentialArgs = {
+  credential: AddDigitalCredential
+}
+
+export type { UpdateCredentialStateArgs } from '@sphereon/ssi-sdk.data-store' // TODO create a local copy?
+
+export interface UniqueDigitalCredential {
+  hash: string
+  digitalCredential: DigitalCredential
```
nklomp, line 109: I would also expect the an optional id value in here. For (V)C this is a mandatory attribute that is unique in the scope of an issuer (so not perse unique in a DB)

nklomp (changes requested):
Some initial remarks

packages/credential-store/src/types/filters.ts
```diff
@@ -7,16 +8,24 @@ import { CredentialRole, DigitalCredential, DocumentType, FindDigitalCredentialA
...
+  const filter: FindDigitalCredentialArgs = [
+    {
+      hash: idOrHash,
+      credentialRole,
+    },
+  ]
+
+  if (uuidValidate(idOrHash)) {
+    filter.push({
+      id: idOrHash,
```
nklomp, line 22: Credential id's certainly do not have to be UUIDs

sanderPostma: Then it is searching the wrong id. The record id is of type uuid
```
    await queryRunner.query(`
      CREATE TABLE "DigitalCredential" (
        "id" uuid NOT NULL DEFAULT uuid_generate_v4(),
```

Do we need a separate field credentialId in the DB? The VC's id is in the document string which can only be searched after parsing the JSON.

sanderPostma: Or the method name is wrong and we do want to search on the PK

packages/credential-store/src/agent/CredentialStore.ts
sanderPostma: Add logger.warning

packages/credential-store/src/types/filters.ts
nklomp: I think both. Being able to search for the id of the credential itself (not pk) makes sense to me,  just like the id (pk) in the db. Probably in the old store they were one and the same

packages/credential-store/src/types/filters.ts
sanderPostma: This is just a helper to avoid duplicate code and make it easier to fetch only VC's
If you need one of the other document types, don't use "verifiableCredentialForRoleFilter"
Create your own filter or create a verifiablePresentationForRoleFilter function when having duplicate code.

packages/credential-store/src/types/ICredentialStore.ts
sanderPostma: fixed in next commit

packages/credential-store/src/types/filters.ts
sanderPostma: I'll have to hoist the credentialId out of the credential to be able to search on it.

sanderPostma: added credentialId: idOrHash above (next commit)

packages/credential-store/src/agent/CredentialStore.ts
```diff
@@ -0,0 +1,257 @@
...
+  GetCredentialsByClaimsArgs,
+  GetCredentialsByIdOrHashArgs,
+  ICredentialStore,
+  OptionalUniqueDigitalCredential,
+  schema,
+  TClaimsColumns,
+  UniqueDigitalCredential,
+} from '../index'
+import { DigitalCredential, UpdateCredentialStateArgs } from '@sphereon/ssi-sdk.data-store'
+import { AbstractDigitalCredentialStore } from '@sphereon/ssi-sdk.data-store/dist/digitalCredential/AbstractDigitalCredentialStore'
```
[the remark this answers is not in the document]
sanderPostma: fixed

packages/oid4vci-holder/src/agent/OID4VCIHolder.ts
sanderPostma: fixed
(The FIXME's were supposed to be in the draft PR to help address them)

packages/credential-store/src/agent/CredentialStore.ts
sanderPostma: fixed

packages/oid4vci-holder/src/agent/OID4VCIHolder.ts
```diff
@@ -634,7 +634,7 @@ export class OID4VCIHolder implements IAgentPlugin {
       })
 
       const credential = {
-        id: issuanceOpt.credentialConfigurationId ?? issuanceOpt.id,
+        id: (issuanceOpt.credentialConfigurationId ?? issuanceOpt.id) as string | undefined, // cast due to error:  Type 'unknown' is not assignable to type 'string | undefined'.
```
nklomp, line 637:
```suggestion
        id: (issuanceOpt.credentialConfigurationId ?? issuanceOpt.id as string | undefined) , // cast due to error:  Type 'unknown' is not assignable to type 'string | undefined'.
```

nklomp: Well the bigger question is why it fails. I know that the lower level libs have some changes. But that has been fixed in develop. The fact develop builds and this branch doesn't suggest either the PR is reverting something, not using the latest develop, or new code is introduced where this happens. Please investigate first before continuing @sanderPostma

nklomp: Ah okay, now I understand what you mean. Yes this was fixed in lates develop

sanderPostma: @nklomp 
Looks like build are passing now indeed

nklomp (approved):

PR #231, opened 2024-08-04
SPRIND-25 - Update the SDK SD-JWT plugin to support non DID use cases

sksadjad:


packages/sd-jwt/src/action-handler.ts
```diff
@@ -70,28 +69,43 @@ export class SDJwtPlugin implements IAgentPlugin {
...
-    if (!key) {
-      throw new Error(`No key found with the given id: ${issuer}`)
+  async getSignKey (
+      identifier: string,
+      context: IRequiredContext
+      //todo make the type of key known
+  ): Promise<{ alg: string; key: any }> {
+    if (identifier.startsWith('did:')) {
+      const didDocument = await context.agent.resolveDid({ didUrl: identifier });
+      const key = didDocument.verificationMethod?.find(vm => vm.publicKeyJwk);
```
nklomp, line 83: This will not work. You need agent Keys not a VM.

We have getFirstKeyWithRelation method, that does the mapping of agent available keys onto a DID  Document.
In this case you would need an additional option to the method to define whether it is an authentication (holder) or assertionMethod (issuer)

packages/sd-jwt/src/action-handler.ts
```diff
@@ -70,28 +69,43 @@ export class SDJwtPlugin implements IAgentPlugin {
...
+      if (!key) {
+        throw new Error(`No verification key found for DID: ${identifier}`);
+      }
+      const alg = key.publicKeyJwk.alg;
+      return { alg, key };
+    } else if (identifier.includes('x5c')) {
+      const x5cKey = await context.agent.getCertificateChain({ identifier });
```
nklomp, line 90: A cetificate chain is far from a key. It is an array of certificates in DER format, which in turn contains a public key.

You will only want to get the public key matched against an IKey of the agent for the certificate that is at the end/or beginning (I always forget the order of the x5c). So you would need to extract the key from the DER (or PEM). I am working on X509 related functionalities. But I rather have a big fixme and error thrown, then this code which isn't remotely what it should do

packages/sd-jwt/src/action-handler.ts
```diff
@@ -70,28 +69,43 @@ export class SDJwtPlugin implements IAgentPlugin {
...
+      if (!x5cKey || !Array.isArray(x5cKey)) {
+        throw new Error(`Invalid x5c header found for identifier: ${identifier}`);
+      }
+      //fixme: change this value to the real alg value
+      const alg = 'RS256';
+      return { alg, key: { x5c: x5cKey } };
+    } else {
+      const key = await context.agent.keyManagerGet({ kid: identifier });
+      const alg = key.meta?.alg;
```
nklomp, line 99: This is too naive. The kid could be either a Veramo keyRef and thus either an alias or publicKeyHex. Or it could be (and more likely) an actual kid in the form of a jwkThumbprint. The kmsLocal implementation stores thumbPrints in the key metadata. Probably we need to do that for the new Musap KMS as well. For now you can test against the KML local implementation

nklomp (changes requested):
Getting the correct identifiers is way more complicated then the current implementation!

nklomp:
Also please first focus on the Verification of a SD-JWT VC. We are getting them from the PID issuer first. We need to get that working ASAP

packages/sd-jwt/src/action-handler.ts
sksadjad: changed the bit that handles the did identifiers and used different functionalities to get to the key object

packages/sd-jwt/src/action-handler.ts
sksadjad: refactored the x5c handling part. I wrote a couple of tests in our crypto-extensions to make sure this will work

packages/sd-jwt/src/action-handler.ts
sksadjad: added functionality for fetching by jwkThumbprints in the crypto-extensions pr and tests for that

packages/sd-jwt/src/action-handler.ts
```diff
@@ -186,8 +188,33 @@ export class SDJwtPlugin implements IAgentPlugin {
...
     const issuer: string = ((decodedVC.jwt as Jwt).payload as Record<string, unknown>).iss as string
     const verifierKey: SignKeyResult = await this.getSignKey({ identifier: issuer, vmRelationship: 'verificationMethod' }, context)
-    const key = await context.agent.keyManagerGet({ kid: verifierKey.key.jwkThumbprint ?? verifierKey.key.kid })
-    return this.algorithms.verifySignature(data, signature, toJwk(key.publicKeyHex, key.type))
+    const x5c: string[] | undefined = (decodedVC.jwt as Jwt).header?.x5c as string[]
+    const x5t: string | undefined = (decodedVC.jwt as Jwt).header?.x5t as string
+    let publicKey: string | undefined
+
+    if (x5c) {
+      const certChain = x5c.map((cert) => `-----BEGIN CERTIFICATE-----\n${cert}\n-----END CERTIFICATE-----`)
```
nklomp, line 196: This method can handle ders and pems. No need to do this manual conversion. Besides we have more robust methods to do that.

packages/sd-jwt/src/action-handler.ts
```diff
@@ -186,8 +188,33 @@ export class SDJwtPlugin implements IAgentPlugin {
...
+      const pemCert = base64ToPEM(x5c[0], 'CERTIFICATE')
+      publicKey = publicKeyHexFromPEM(pemCert)
+      const isValidCertChain = await context.agent.verifyCertificateChain({ chain: certChain, trustAnchors: [funkeTestCA, funkeTestIssuer] })
```
nklomp, line 199: Why are you adding a non CA as a trust anchor? The FunkeTestIssuer certificate is the first in the actual chain and has been issued by the funcke CA. So only the Funke CA and our CA should be in here. With a very big FIXME, that is should be configurable anyway

packages/sd-jwt/src/action-handler.ts
```diff
@@ -186,8 +188,33 @@ export class SDJwtPlugin implements IAgentPlugin {
...
+
+      if (isValidCertChain.error) {
+        throw new Error('Certificate chain validation failed')
+      }
+    } else if (x5t) {
+      publicKey = x5t
```
nklomp, line 205: Why are you doing this. This is the hash of the cert. The code above should have been getting the first cert and its public key, which the verifyCertChain returns

packages/sd-jwt/src/action-handler.ts
```diff
@@ -186,8 +188,33 @@ export class SDJwtPlugin implements IAgentPlugin {
...
+    } else {
+      const verifierKey: SignKeyResult = await this.getSignKey({ identifier: issuer, vmRelationship: 'verificationMethod' }, context)
+      const key = await context.agent.keyManagerGet({ kid: verifierKey.key.jwkThumbprint ?? verifierKey.key.kid })
+      publicKey = key.publicKeyHex
+    }
+
+    if (!publicKey) {
+      throw new Error('No valid public key found for signature verification')
+    }
+    publicKey = publicKey ?? verifierKey.key.jwkThumbprint ?? verifierKey.key.kid
```
nklomp, line 215: You are throwing an error above this line, so the first part of the expression always contains a value and thus the others never come into play to begin with

packages/sd-jwt/src/action-handler.ts
```diff
@@ -175,22 +187,34 @@ export class SDJwtPlugin implements IAgentPlugin {
   async verify(sdjwt: SDJwtVcInstance, context: IRequiredContext, data: string, signature: string) {
     const decodedVC = await sdjwt.decode(`${data}.${signature}`)
```
nklomp, line 188: Why are you involving the KMS on the verify function? The verify happens 99% of the time on some-one elses SD-JWT. And even if it were our own, we still should not involve the KMS for verification. The KMS is only involved in issuing/signing/presenting

packages/sd-jwt/src/action-handler.ts
sksadjad: changed this to just pass in the first one of the chain

packages/sd-jwt/src/action-handler.ts
sksadjad: removed this part related to the `x5t`. now I'm just looking at two things: issuers if they are did, certificate chains that can be found in the `x5c` header param.
Then I'm calling into the function `x5cToPemCertChain` and `PEMToJwk` from `x509-utils` to get the json web key

packages/sd-jwt/src/action-handler.ts
```diff
@@ -186,8 +188,33 @@ export class SDJwtPlugin implements IAgentPlugin {
...
+      const isValidCertChain = await context.agent.verifyCertificateChain({ chain: certChain, trustAnchors: [funkeTestCA, funkeTestIssuer] })
+
+      if (isValidCertChain.error) {
+        throw new Error('Certificate chain validation failed')
+      }
+    } else if (x5t) {
+      publicKey = x5t
+    } else {
+      const verifierKey: SignKeyResult = await this.getSignKey({ identifier: issuer, vmRelationship: 'verificationMethod' }, context)
+      const key = await context.agent.keyManagerGet({ kid: verifierKey.key.jwkThumbprint ?? verifierKey.key.kid })
```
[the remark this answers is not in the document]
sksadjad: removed the KMS dependency as we're not using that for resolving the public key

packages/sd-jwt/src/action-handler.ts
sksadjad: rewrote this part. now we're getting `jwk` from the did or from the `x5c`

packages/sd-jwt/src/action-handler.ts
sksadjad: yeah. correct. I have removed the kms calls from this part of the code which is responsible for verifying and added functions for getting the `JWK` from certificate chain of the issuer did. this way we're not calling into our KMS for fetching a public key

packages/sd-jwt/src/action-handler.ts
```diff
@@ -175,22 +190,37 @@ export class SDJwtPlugin implements IAgentPlugin {
...
     const issuer: string = ((decodedVC.jwt as Jwt).payload as Record<string, unknown>).iss as string
-    if (!issuer.startsWith('did:')) {
-      throw new Error('invalid_issuer: issuer must be a did')
+    const header = (decodedVC.jwt as Jwt).header as Record<string, any>
+    const x5c: string[] | undefined = header?.x5c as string[]
+    let jwk: JWK | JsonWebKey | undefined = undefined
+    let key: IKey | undefined
+    if (issuer.includes('did:')) {
+      const identifier: IIdentifier = await context.agent.didManagerGet({ did: issuer.split('#')[0] })
+      key = await context.agent.getKey({ identifier })
```
nklomp, line 199: Huh? why are we asking the agent again for an identifier it controls? We should be doing DID resolution for the DID and then we should be matching the kid from the SD-JWT header to get the correct VM and thus public key

packages/sd-jwt/src/action-handler.ts
```diff
@@ -175,22 +190,37 @@ export class SDJwtPlugin implements IAgentPlugin {
...
+      const firstCert = x5c[0]
+      const certificateValidationResult: X509ValidationResult = await context.agent.verifyCertificateChain({
+        chain: [firstCert],
+        trustAnchors: [funkeTestCA, sphereonCA],
+      })
+
+      if (certificateValidationResult.error || !certificateValidationResult?.certificateChain) {
+        throw new Error('Certificate chain validation failed')
+      }
+      const pem = await x5cToPemCertChain([firstCert])
```
nklomp, line 216: The above verify method already returns the JWK for you for every certificate if there are no errors

packages/sd-jwt/src/action-handler.ts
nklomp: The code you have been replacing was literally doing that btw

sksadjad: reverted the code that was doing the did resolution

packages/sd-jwt/src/action-handler.ts
sksadjad: replaced it with getting the jwk value from the first element of the certificate chain validation result

PR #232, opened 2024-08-12
integrating support for SD-JWT into an RP setup

sksadjad:
Updated the modules:
  - ebsi-support (minor change)
  - presentation-exchange (minor change)
  - sd-jwt (some type change to be inline with the reference lib)
  - siopv2-oid4vp-common (minor change)
  - siopv2-oid4vp-op-auth
    - updated to use the new did-auth-siop
    - implemented and passed the new callback functions
    - changed it to work with none-did identifiers
    - changed it to use the new `identifier-resolution` module
    - changed it to use the new `jwt-service` module
  - siopv2-oid4vp-rp-auth
    - updated to use the new did-auth-siop
    - implemented and passed the new callback functions
    - changed it to work with none-did identifiers
    - changed it to use the new `identifier-resolution` module
    - changed it to use the new `jwt-service` module
  - siopv2-oid4vp-rp-rest-api (minor change)
  - siopv2-oid4vp-rp-rest-client (minor change)


packages/siopv2-oid4vp-rp-rest-api/__tests__/agent.ts
```diff
@@ -27,6 +27,7 @@ import { IPresentationExchange, PresentationExchange } from '@sphereon/ssi-sdk.p
 import { CheckLinkedDomain } from '@sphereon/did-auth-siop'
 import { entraAndSphereonCompatibleDef, entraVerifiedIdPresentation } from './presentationDefinitions'
 import Debug from 'debug'
+import { createHash } from 'crypto'
```
nklomp, line 30: We already should have this in the SD-JWT plugin right?. Also since this is the REST API on top it should use the other oid4vp base plugin anyway right? So then I rather see this is an optional param, with a default in the siopv2-oid4vp-rp-auth SSI-SDK plugin

packages/siopv2-oid4vp-rp-auth/src/agent/SIOPv2RP.ts
```diff
@@ -91,21 +97,25 @@ export class SIOPv2RP implements IAgentPlugin {
...
     if (
       responseState.status === AuthorizationResponseStateStatus.VERIFIED &&
       args.includeVerifiedData &&
       args.includeVerifiedData !== VerifiedDataMode.NONE
     ) {
+      // todo this should also include mdl-mdoc
       const presentationDecoded = CredentialMapper.decodeVerifiablePresentation(
         responseState.response.payload.vp_token as OriginalVerifiablePresentation,
+        //todo: later we want to conditionally pass in options for mdl-mdoc here
+        rpInstance.rpOptions.credentialOpts?.hasher,
```
nklomp, line 118: Does the mapper provide a default hasher? Otherwise we might want to provide it here

packages/siopv2-oid4vp-rp-rest-api/__tests__/agent.ts
nklomp: Also is this the same hasher we use in the SD-JWT plugin to begin with?

packages/siopv2-oid4vp-op-auth/src/session/OID4VP.ts
```diff
@@ -159,16 +159,16 @@ export class OID4VP {
...
       verifiableCredentials: vcs.credentials,
       allDIDs: this.allDIDs,
       hasher: opts?.hasher,
     }).createVerifiablePresentation(vcs.definition.definition, vcs.credentials, signCallback, {
       proofOptions,
       // fixme: Update to newer siop-vp to not require dids here.
-
-      holderDID: isManagedIdentifierDidOpts(idOpts) ? (await this.session.context.agent.identifierManagedGetByDid(idOpts)).did : undefined,
+      // holderDID: isManagedIdentifierDidOpts(idOpts) ? (await this.session.context.agent.identifierManagedGetByDid(idOpts)).did : undefined,
+      holderDID: identifier.kid,
```
nklomp, line 171: That should not be the fix for the above code. The code before was explicitly only setting the holderDID when it actually was a DID. The fixme was about the fact that we seem to rely on DIDs. So without having looked at it the 'holderDID' property should probably change.

packages/siopv2-oid4vp-op-auth/src/session/functions.ts
```diff
@@ -130,6 +124,27 @@ export async function createOPBuilder({
...
+    checkLinkedDomain: 'never' | 'if_present' | 'always'
+    wellknownDIDVerifyCallback?: VerifyCallback
+  },
+): VerifyJwtCallback {
+  return async (jwtVerifier, jwt) => {
+    resolver = resolver ?? getResolver({ subjectSyntaxTypesSupported: ['ethr', 'ion'] })
+    const audience =
+      jwtVerifier.type === 'request-object'
+        ? (verifyOpts?.audience ?? getAudience(jwt.raw))
+        : jwtVerifier.type === 'id-token'
```
nklomp, line 139: This is a really odd ternary that could just be a single ternary with an or for the type

packages/siopv2-oid4vp-rp-auth/src/functions.ts
```diff
@@ -98,18 +103,22 @@ export async function createRPBuilder(args: {
   }
 
   const resolution = await context.agent.identifierManagedGet(didOpts.idOpts)
+  const resolver =
```
nklomp, line 106: Where is this resolver passed to the builder like it did in the past?

packages/siopv2-oid4vp-rp-auth/src/functions.ts
```diff
@@ -118,39 +127,71 @@ export async function createRPBuilder(args: {
...
+    resolver = resolver ?? getResolver({ subjectSyntaxTypesSupported: ['ethr', 'ion'] })
+    const audience =
+      jwtVerifier.type === 'request-object'
+        ? (verifyOpts?.audience ?? getAudience(jwt.raw))
+        : jwtVerifier.type === 'id-token'
+          ? (verifyOpts?.audience ?? getAudience(jwt.raw))
+          : undefined
+
+    await verifyDidJWT(jwt.raw, resolver, { audience, ...verifyOpts })
+    return true
```
nklomp, line 191: Please add a fixme if you are creating code like this. If this gets overlooked we are simply not actually validating anymore

packages/siopv2-oid4vp-rp-auth/src/functions.ts
sksadjad: the `resolver` param is not expected in the builder any more. I'm just passing it to the verification callback function

packages/siopv2-oid4vp-op-auth/src/session/OID4VP.ts
sksadjad: `holderDID` here is a term that is used in the pex only and as far as I understand doesn't have to be a did. it can be easily refactored to be just `holder`. therefore I don't think we should enforce it being a did. so in the OP side, I think, we can use the `kid` of a managed identifier.

packages/siopv2-oid4vp-op-auth/src/session/functions.ts
sksadjad: simplified the ternary function.

packages/siopv2-oid4vp-rp-auth/src/agent/SIOPv2RP.ts
sksadjad: added the default hasher

packages/siopv2-oid4vp-rp-rest-api/__tests__/agent.ts
sksadjad: this is a test file and we're using the `createHash` from `crypto` in there too

packages/siopv2-oid4vp-op-auth/src/session/functions.ts
```diff
@@ -130,6 +127,79 @@ export async function createOPBuilder({
...
 
+//fixme: this is written based on OID4VCIHolder.signCallback sync the fixes to this with that function
+export function signCallback<T extends JwtIssuer & JwtIssuanceContextBase>(
+  idOpts: ManagedIdentifierOpts,
+  context: IRequiredContext,
+): (jwt: { header: JwtHeader; payload: JwtPayload }, kid?: string) => Promise<string> {
+  return async (jwt: { header: JwtHeader; payload: JwtPayload }, kid?: string) => {
+    let iss = jwt.payload.iss
+    const jwk = jwt.header.jwk
+
```
nklomp, line 138: You should not have to do anything of the VCI flow here. The original signer get its arguments pretty much as it should be, so you basically should call the compactJwsSign function. and not the DIDSign. That is the whole point of what you should change

packages/siopv2-oid4vp-rp-auth/src/functions.ts
```diff
@@ -118,39 +134,127 @@ export async function createRPBuilder(args: {
...
+    }
+
+    const resolution = await context.agent.identifierManagedGet(idOpts)
+    if (!iss && isManagedIdentifierDidResult(resolution)) {
+      iss = resolution.did
+    } else {
+      iss = resolution.issuer
+    }
+    if (!iss) {
+      return Promise.reject(Error(`No issuer could be determined from the JWT ${JSON.stringify(jwt)}`))
```
nklomp, line 215: See remark for the other signer. Please just call the compact jwt signer. Most of the rest is not needed here

packages/siopv2-oid4vp-rp-auth/src/functions.ts
sksadjad: added a todo for both places that we're calling the verify function

packages/siopv2-oid4vp-op-auth/src/session/functions.ts
sksadjad: refactored the function to use `jwtCreateJwsCompactSignature` function

packages/siopv2-oid4vp-rp-auth/src/functions.ts
sksadjad: refactored the function to use `jwtCreateJwsCompactSignature` function

packages/siopv2-oid4vp-op-auth/src/session/OpSession.ts
```diff
@@ -221,12 +220,13 @@ export class OpSession {
     if (!resolveOpts.subjectSyntaxTypesSupported || resolveOpts.subjectSyntaxTypesSupported.length === 0) {
       resolveOpts.subjectSyntaxTypesSupported = await this.getSupportedDIDMethods(true)
     }
+    //todo: populate with the right verification params. In did-auth-siop we don't have any test that actually passes this parameter
+    // and furthermore, it receives a couple of parameters that don't make that much sense for OP. like presentationVerificationCallback and revocationOpts
```
nklomp, line 224: This comment doesn't make any sense, as linked domains need to be verified on both ends

packages/siopv2-oid4vp-op-auth/src/session/OpSession.ts
```diff
@@ -221,12 +220,13 @@ export class OpSession {
...
     const verification: Verification = {
-      mode: VerificationMode.INTERNAL,
-      checkLinkedDomain: CheckLinkedDomain.IF_PRESENT,
-      resolveOpts,
+      // mode: VerificationMode.INTERNAL,
```
nklomp, line 226: Why is this disabled, as the default is an opt in anyway

packages/siopv2-oid4vp-op-auth/src/session/functions.ts
nklomp: Why is all this code back again?

packages/siopv2-oid4vp-op-auth/src/session/functions.ts
```diff
@@ -130,6 +128,71 @@ export async function createOPBuilder({
...
+    checkLinkedDomain: 'never' | 'if_present' | 'always'
+    wellknownDIDVerifyCallback?: VerifyCallback
+  },
+): VerifyJwtCallback {
+  return async (jwtVerifier, jwt) => {
+    resolver = resolver ?? getResolver({ subjectSyntaxTypesSupported: ['ethr', 'ion'] })
+    const audience =
+      jwtVerifier.type === 'request-object' || jwtVerifier.type === 'id-token' ? (verifyOpts?.audience ?? getAudience(jwt.raw)) : undefined
+    //todo probably wise to revisit this. this is called verifyDidJWT and expects a did resolver param.
+    await verifyDidJWT(jwt.raw, resolver, { audience, ...verifyOpts })
```
nklomp, line 191: Why is this calling verifyDidJwt instead of the new signature service?

packages/siopv2-oid4vp-rp-auth/src/functions.ts
nklomp: What is up with the code being reverted to the old state again, depending on the verifyDidJwt instead of the signature service :(

packages/siopv2-oid4vp-op-auth/src/session/OpSession.ts
sksadjad: removed the second part of the comment (`and furthermore...`).
the Verification object is now:
```
export interface Verification {
  presentationVerificationCallback?: PresentationVerificationCallback
  revocationOpts?: RevocationOpts
  replayRegistry?: IRPSessionManager
}
```

packages/siopv2-oid4vp-op-auth/src/session/OpSession.ts
sksadjad: the verification object has been changed:
```
export interface Verification {
  presentationVerificationCallback?: PresentationVerificationCallback
  revocationOpts?: RevocationOpts
  replayRegistry?: IRPSessionManager
}
```
and the todo that I've added is talking about populating it with the correct values

packages/siopv2-oid4vp-op-auth/src/session/functions.ts
sksadjad: not sure what your question is. code has been refactored and now I am using the `identifierManagedGet` to get a `resolution` for my `idOpts` and at the end calling `jwtCreateJwsCompactSignature` for signing the payload.
```
    const jwk = jwt.header.jwk

    const resolution = await context.agent.identifierManagedGet(idOpts)
    const issuer = jwt.payload.iss || (isManagedIdentifierDidResult(resolution) ? resolution.did : resolution.issuer)
    if (!issuer) {
      return Promise.reject(Error(`No issuer could be determined from the JWT ${JSON.stringify(jwt)}`))
    }
    kid = resolution.kid
    if (kid && isManagedIdentifierDidResult(resolution) && !kid.startsWith(resolution.did)) {
      // Make sure we create a fully qualified kid
      const hash = kid.startsWith('#') ? '' : '#'
      kid = `${resolution.did}${hash}${kid}`
    }
    const header = { ...jwt.header, ...(kid && !jwk && { kid }) } as Partial<JWTHeader>
    const payload = { ...jwt.payload, ...(issuer && { iss: issuer }) }
    if (jwk && header.kid) {
      delete header.kid
    }

    const result: JwsCompactResult = await context.agent.jwtCreateJwsCompactSignature({
      issuer: { identifier: issuer, noIdentifierInHeader: false },
      protectedHeader: header,
      payload,
    })
    return result.jwt
```

packages/siopv2-oid4vp-rp-auth/src/functions.ts
sksadjad: discussed this, for now returning `true` and added a todo with the ticket number for refactoring it

packages/siopv2-oid4vp-op-auth/src/session/functions.ts
sksadjad: discussed this, for now returning true and added a todo with the ticket number for refactoring it

packages/siopv2-oid4vp-rp-auth/src/functions.ts
nklomp: Haven't looked. But not validating is a FIXME of course not a TODO

packages/siopv2-oid4vp-op-auth/src/session/functions.ts
nklomp: Yes but I mentioned that you should not copy the kid stuff from elsewhere. That was once removed, now it is back

packages/siopv2-oid4vp-rp-auth/src/functions.ts
nklomp: Okay apparently it is marked as fixme. Please be explicit when talking about something; as this only provokes unnecessary discussion and work

sksadjad: I've misspoken. it is in fact a fixme

packages/vc-status-list-issuer-rest-api/package.json
```diff
@@ -13,8 +13,8 @@
   },
   "dependencies": {
     "@sphereon/ssi-express-support": "workspace:*",
-    "@sphereon/ssi-sdk-ext.did-utils": "0.24.1-unstable.82",
-    "@sphereon/ssi-sdk-ext.identifier-resolution": "0.24.1-unstable.82",
+    "@sphereon/ssi-sdk-ext.did-utils": "0.24.1-next.96",
+    "@sphereon/ssi-sdk-ext.identifier-resolution": "0.24.1-next.96",
```
nklomp, line 17: Please do no update individual packages. Every ssi-sdk-(ext) always has to be in sync in the whole project.

sksadjad: reverted this (and a couple of other) version updates

PR #238, opened 2024-09-13
feature/SDK-30_default-pd-format

sanderPostma:


packages/siopv2-oid4vp-op-auth/src/session/OID4VP.ts
```diff
@@ -156,7 +156,7 @@ export class OID4VP {
       context: this.session.context,
       domain: proofOptions.domain,
       challenge: proofOptions.challenge,
-      format: opts?.restrictToFormats ?? selectedVerifiableCredentials.definition.definition.format,
+      fallbackFormat: opts?.restrictToFormats ?? selectedVerifiableCredentials.definition.definition.format,
```
nklomp, line 159: This is no fallback format

packages/siopv2-oid4vp-op-auth/src/session/functions.ts
```diff
@@ -47,7 +47,7 @@ export async function createOID4VPPresentationSignCallback({
       fetchRemoteContexts,
       domain,
       challenge,
-      format,
+      fallbackFormat: fallbackFormat,
```
nklomp, line 50: Format not fallback

packages/siopv2-oid4vp-op-auth/src/session/OID4VP.ts
sanderPostma: I'll rename it back, but it's the last option in 
```
 const formatOptions = format ?? presentationDefinition.format ?? args.fallbackFormat
```

nklomp: Yes which results in a format being used afterwards that is not a fallback in the end of course

packages/siopv2-oid4vp-op-auth/src/session/functions.ts
```diff
@@ -34,7 +34,7 @@ export async function createOID4VPPresentationSignCallback({
   challenge?: string
   fetchRemoteContexts?: boolean
   skipDidResolution?: boolean
-  format?: Format
+  fallbackFormat?: Format
```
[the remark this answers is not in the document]
sanderPostma: fixed

packages/siopv2-oid4vp-op-auth/src/session/functions.ts
sanderPostma: reverted

sanderPostma:
@nklomp 
Shouldn't we select a format from opSession.authenticationRequest.vp_formats ? When it supports vc+sd-jwt we can take select that, otherwise another one.

nklomp:
Yes that is correct. But we should already be doing that I think.

PR #241, opened 2024-09-18
Feature/from funke

sanderPostma:
NOT FINISHED


packages/sd-jwt/src/action-handler.ts
```diff
@@ -225,10 +225,29 @@ export class SDJwtPlugin implements IAgentPlugin {
...
-    return this.verifySignatureCallback(context)(data, signature, key)
+    return this.verifySignatureCallback(context)(data, signature, this.getJwk(payload))
   }
 
+  private getJwk(payload: JwtPayload): JsonWebKey {
+    if (payload.cnf?.jwk !== undefined) {
+      return payload.cnf.jwk as JsonWebKey
+    } else if (payload.cnf !== undefined && 'kid' in payload.cnf && typeof payload.cnf.kid === 'string' && payload.cnf.kid.startsWith('did:jwk:')) {
+      // extract JWK from kid FIXME isn't there a did function for this already? Otherwise create one
+      const encoded = this.extractBase64FromDID(payload.cnf.kid)
```
nklomp, line 236: We already have low level support converting any VM to a JWK. So this method could also deal with more than did:jwk: as that is way too naive anyway. The chances of a CNF being a did:jwk only are pretty slim

packages/sd-jwt/src/action-handler.ts
```diff
@@ -225,10 +225,29 @@ export class SDJwtPlugin implements IAgentPlugin {
...
+      const decoded = decodeBase64url(encoded)
+      const jwt = JSON.parse(decoded)
+      return jwt as JsonWebKey
+    }
+    throw Error('Unable to extract JWK from SD-JWT payload')
+  }
+
+  private extractBase64FromDID(did: string): string {
+    const parts = did.split(':')
```
nklomp, line 245: Wait to naive. This is only true for did:jwk

packages/sd-jwt/src/action-handler.ts
```diff
@@ -225,10 +225,29 @@ export class SDJwtPlugin implements IAgentPlugin {
...
+    if (parts.length < 3) {
+      throw new Error('Invalid DID format')
```
nklomp, line 247: No this whole method should really not be used. This has nothing to do with DIDs in general. This is a did:jwk assumption, for which this handling is already odd to begin with

packages/sd-jwt/src/action-handler.ts
nklomp: https://github.com/Sphereon-Opensource/SSI-SDK-crypto-extensions/blob/c0ca69fe0f10cfd9cdafa94b7af31a6cf6100680/packages/did-utils/src/did-functions.ts#L293C4-L294C1

packages/sd-jwt/src/action-handler.ts
sanderPostma: It helped getting the sd-jwt verified for now. We do not supply cnf.jwk and without out the sd-jwt verification fails

packages/sd-jwt/src/action-handler.ts
nklomp: https://github.com/Sphereon-Opensource/SSI-SDK-crypto-extensions/blob/c0ca69fe0f10cfd9cdafa94b7af31a6cf6100680/packages/did-utils/src/did-functions.ts#L451

nklomp: And many other functions you can use in that package. When dealing with DIDs look into the did-utils module. When dealing with key conversions look into the key-utils module. Code like the above should never live in the SD-JWT module

nklomp (commented):
See remark

PR #242, opened 2024-09-18
fix/SPRIND-60

BtencateSphereon:


packages/siopv2-oid4vp-op-auth/src/session/OID4VP.ts
```diff
@@ -169,7 +175,8 @@ export class OID4VP {
       format: opts?.restrictToFormats ?? selectedVerifiableCredentials.definition.definition.format,
       skipDidResolution: opts?.skipDidResolution ?? false,
     })
-    const identifier: ManagedIdentifierResult = await this.session.context.agent.identifierManagedGet(idOpts)
+    const identifier: ManagedIdentifierResult | undefined = idOpts.method !== 'oid4vci-issuer' ? await this.session.context.agent.identifierManagedGet(idOpts) : undefined
```
nklomp, line 178: Isn't vci-issuer returning a MaangedIdentifierResult?

nklomp (changes requested):
I would expect a managed identifier result to happen for a oid4vci issuer as well

packages/siopv2-oid4vp-op-auth/src/session/OID4VP.ts
BtencateSphereon: it is, so we can call it anyways changed

nklomp (approved):

PR #243, opened 2024-09-23
CWALL-254: Added vc validation against the schema

sksadjad:
IMPORTANT NOTE
the examples that I see from ebsi, all have some discrepancies. so I've added an option to dictate whether we want to check the vc against it's schema or not:

```
export type VerifyCredentialToAcceptArgs = {
  mappedCredential: MappedCredentialToAccept
  hasher?: Hasher
  verifyAgainstSchema?: boolean // newly added value
  context: RequiredContext
}
```


nklomp:
The logic should be the other way around. Just like how we handle expiration and credential status. When the object is present, it means the issuer values the respective feature of the VCDM. So when they issue with a schema, we should always validate, unless an option was set to not validate. So something like noSchemaValidation makes more sense. Or even nicer an enum named SchemaValidation, with values, ALWAYS, NEVER, WHEN_PRESENT. That allows you to configure from a wallet or RP perspective whether you always need schema's, whether you will not verify schema's or only when the issuer added schema's

PR #247, opened 2024-10-02
feature/SPRIND-54

sksadjad:
add:
- Activity logs
modified:
- event log entity to have activity log properties
- added discriminator value for type of log in the db
- queries for select. the previous query was selecting **OR** (in case of multiple criterias) and now, I've modified it to be **AND**
fix:
- fixed deleting **digitalCredential** in case of it having child credentials


nklomp (approved):
LGTM

PR #255, opened 2024-10-18
feature/SPRIND-84

zoemaas:


packages/oidf-client/src/agent/OIDFClient.ts
```diff
@@ -0,0 +1,51 @@
...
+        verifyJwt: this.verifyJwt.bind(this)
+    }
+
+    private async resolveTrustChain(args: ResolveTrustChainArgs): Promise<ResolveTrustChainCallbackResult> {
+        const { entityIdentifier, trustAnchors } = args
+        return this.oidfClient.resolveTrustChain(entityIdentifier, trustAnchors)
+    }
+
+    private async registerCryptoServiceCallback(platformCallback: RegisterCryptoServiceCallbackArgs): Promise<RegisterCryptoServiceCallbackResult> {
+        CryptoServiceJS.register(platformCallback);
```
nklomp, line 44: Please do not expose this function in the SDK. We have a jwt-service that should be used

zoemaas: Removed the function from the SDK. As far as I can understand, the default service should be used. Is that correct? Does it make sense to keep the verifyJwt() function exposed?

nklomp: We have a plugin that does JWTs/JWS. So that should be leveraged to sign JWTs, as it has integrated support for all the identifier types we support (X509, DIDs, JWKs, kids etc). 

The verify function is also exposed in the OIDF client. Depending on whether that function is doing more than a regular JWT check it makes sense to delegate it to the OIDF client. IF not then it could also use the JWT exposed CompactJwt/Jwe verify function from the SDK.

What would be nice if you have the option to have a callback. But then it should not be exposed as a method, but as a constructor argument to the plugin. Then you could do something like:

If callback is set on the plugin, delegate to the respective function of the callback. If not then delegate to the JWT service of the SDK

zoemaas: Added the verifyJwt() and signJwt functions using `@sphereon/ssi-sdk-ext.jwt-service`.  Still cannot make it work properly because the `CryptoService.register(...)` function from the OIDF client is ignoring the input.  I'm checking it with @jcmelati

zoemaas: Added the verifyJwt() and signJwt() functions using `@sphereon/ssi-sdk-ext.jwt-service`.  Still cannot make it work properly because the second parameter of the verifyJwt function (context). I get a compilation error which I'm not sure how to fix.

packages/oidf-client/package.json
```diff
@@ -0,0 +1,52 @@
...
+			"IOIDFClient": "./src/types/IOIDFClient.ts"
+		}
+	},
+	"scripts": {
+		"build": "tsc",
+		"build:clean": "tsc --build --clean && tsc --build",
+		"generate-plugin-schema": "ts-node ../../packages/dev/bin/sphereon.js dev generate-plugin-schema"
+	},
+	"dependencies": {
+		"openid-federation-modules-openapi": "link:../../../OpenID-Federation/build/js/packages/openid-federation-modules-openapi",
```
sanderPostma, line 18: @zoemaas 
Please use public packages now
https://www.npmjs.com/search?q=%40Sphereon%2FOpenID-Federation

sanderPostma (changes requested):

packages/oidf-client/package.json
zoemaas: Fixed the issues with the public packages

packages/oidf-client/__tests__/shared/oidfClientAgentLogic.ts
```diff
@@ -0,0 +1,82 @@
...
+    afterAll(testContext.tearDown)
+
+    it('should build a trust chain 1', async () => {
+      nockSetup()
+      const trustChain = await agent.resolveTrustChain({
+        entityIdentifier: "https://spid.wbss.it/Spid/oidc/rp/ipasv_lt",
+        trustAnchors:["https://oidc.registry.servizicie.interno.gov.it"]
+      })
+
+      expect(trustChain).toBeDefined()
```
BtencateSphereon, line 54: this is going to fail if the result is null, as null is a defined value and the result of agent.resolveTrustChain can be null. 

Not sure why the result of agent.resolveTrustChain is so complex anyways

BtencateSphereon (changes requested):

packages/oidf-client/__tests__/shared/oidfClientAgentLogic.ts
zoemaas: Fixed the non-null expectation of the tests. 

I cannot answer the second question, I'm just using the logic that is already there. It's not up to me to decide.

PR #259, opened 2024-10-29
feature/SPRIND-92

BtencateSphereon:


packages/data-store/src/__tests__/contact.store.test.ts
```diff
@@ -1208,15 +1217,29 @@ describe('Contact store tests', (): void => {
             correlationId: 'example_did3',
           },
         },
+        {
+          alias: 'test_alias4',
+          origin: IdentityOrigin.EXTERNAL,
+          roles: [CredentialRole.FEDERATION_OPERATOR],
+          identifier: {
+            type: CorrelationIdentifierType.DID,
```
nklomp, line 1225: Url not did for federation

packages/data-store/src/__tests__/contact.store.test.ts
```diff
@@ -1208,15 +1217,29 @@ describe('Contact store tests', (): void => {
             correlationId: 'example_did3',
           },
         },
+        {
+          alias: 'test_alias4',
+          origin: IdentityOrigin.EXTERNAL,
+          roles: [CredentialRole.FEDERATION_OPERATOR],
```
nklomp, line 1223: Probably we should rename this to FEDERATION_TRUST_ANCHOR to be more inline with the spec , as you can have your roots of trust anywhere in the federation.  Doesn't necessarily have to be at the topmost level with the operator. For now I am okay with calling it operator in the UI, but in our datamodel it should use the correct term

packages/oid4vci-holder/src/agent/OID4VCIHolder.ts
```diff
@@ -1016,6 +1008,34 @@ export class OID4VCIHolder implements IAgentPlugin {
     logger.log(`Notification to ${notificationEndpoint} has been dispatched`)
   }
 
+  private async oid4vciGetFederationTrust(args: GetFederationTrustArgs, context: RequiredContext): Promise<boolean> {
+    const { requestData, serverMetadata, trustAnchors } = args
```
nklomp, line 1012: This method should live in the federation client module probably. If we provide an Issuer string argument there would be nothing left that is specific for oid4vc an we could simple call the method from here using the servermetadata issuer as argument value

packages/oid4vci-holder/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@sphereon/ssi-sdk.oid4vci-holder",
-  "version": "0.30.1",
+  "version": "0.30.99-bram",
```
sanderPostma, line 3: Where is the PR for this version? It needs to be reviewed & merged first so this can be updated.

nklomp (changes requested):
Some small remarks

packages/oid4vci-holder/src/agent/OID4VCIHolder.ts
```diff
@@ -215,6 +216,7 @@ export class OID4VCIHolder implements IAgentPlugin {
     oid4vciHolderStoreCredentialBranding: this.oid4vciHolderStoreCredentialBranding.bind(this),
     oid4vciHolderStoreCredentials: this.oid4vciHolderStoreCredentials.bind(this),
     oid4vciHolderSendNotification: this.oid4vciHolderSendNotification.bind(this),
+    oid4vciGetFederationTrust: this.oid4vciGetFederationTrust.bind(this),
```
sanderPostma, line 219: Was this supposed to go into oid4vci-holder?
Quote from Niels in my PR:
> OIDF has nothing to do with OID4VC, except for the fact it can be used as a registry for trust establishment. So it for sure does not belong in the VCI REST API or anything

packages/data-store/src/types/digitalCredential/digitalCredential.ts
```diff
@@ -93,6 +93,7 @@ export enum CredentialRole {
   ISSUER = 'ISSUER',
   VERIFIER = 'VERIFIER',
   HOLDER = 'HOLDER',
+  FEDERATION_OPERATOR = 'FEDERATION OPERATOR'
```
sanderPostma, line 96: FEDERATION_OPERATOR instead of FEDERATION OPERATOR (also in the SQL above,  This might cause issues in some SQL implementations.)

sanderPostma: I also understood that enum keys & values should be the same to allow casting.

packages/oid4vci-holder/src/agent/OID4VCIHolder.ts
```diff
@@ -1016,6 +1008,34 @@ export class OID4VCIHolder implements IAgentPlugin {
...
+    const params = new URLSearchParams(url.search)
+    const openidFederation = params.get('openid_federation')
+
+    const entityIdentifier = openidFederation ?? serverMetadata.issuer
+
+    return context.agent.resolveTrustChain({
+      entityIdentifier,
+      trustAnchors
+    })
+    .then((result) => !result || result.length === 0)
```
sanderPostma, line 1036: What is result of resolveTrustChain? (I cannot see the type also) 
This looks like you return true when no trust chain was found, so returning true when not trusted. I may be wrong, but I can't see that clearly.

packages/oid4vci-holder/src/machine/oid4vciMachine.ts
```diff
@@ -245,6 +261,10 @@ const createOID4VCIMachine = (opts?: CreateOID4VCIMachineOpts): OID4VCIStateMach
       [OID4VCIMachineStates.transitionFromSetup]: {
         id: OID4VCIMachineStates.transitionFromSetup,
         always: [
+          {
+            target: OID4VCIMachineStates.getFederationTrust,
+            cond: OID4VCIMachineGuards.oid4vciIsOIDFOriginGuard && OID4VCIMachineGuards.oid4vciHasTrustedAnchors,
```
sanderPostma, line 266:
```
 cond: [OID4VCIMachineGuards.oid4vciIsOIDFOriginGuard , OID4VCIMachineGuards.oid4vciHasTrustedAnchors]
 ```
 Now you are &&ing two strings

File: packages/oid4vci-holder/src/machine/oid4vciMachine.ts [hunk 9]
```diff
@@ -271,6 +291,28 @@ const createOID4VCIMachine = (opts?: CreateOID4VCIMachineOpts): OID4VCIStateMach
...
+            target: OID4VCIMachineStates.addContact,
+            actions: assign({
+              isTrustedFederation: (_ctx: OID4VCIMachineContext, _event: DoneInvokeEvent<boolean>) => _event.data
+            })
+          },
+          onError: {
+            target: OID4VCIMachineStates.handleError,
+            actions: assign({
+              error: (_ctx: OID4VCIMachineContext, _event: DoneInvokeEvent<Error>): ErrorDetails => ({
+                title: translate('oid4vci_machine_store_credential_error_title'), // TODO
```
sanderPostma, line 308: TODO what?

sanderPostma (changes requested):

packages/oid4vci-holder/src/agent/OID4VCIHolder.ts
nklomp: Yes should be integrated here. No the function should not be exposed to the outside world

packages/oid4vci-holder/src/machine/oid4vciMachine.ts
BtencateSphereon: which seem to work

packages/oid4vci-holder/src/machine/oid4vciMachine.ts
BtencateSphereon: had to set the correct localization string. fixed

packages/data-store/src/types/digitalCredential/digitalCredential.ts
BtencateSphereon: ye changed it

packages/oid4vci-holder/package.json
BtencateSphereon: local testing, reverted

packages/data-store/src/__tests__/contact.store.test.ts
BtencateSphereon: changed to FEDERATION_TRUST_ANCHOR

packages/data-store/src/__tests__/contact.store.test.ts
BtencateSphereon: changed to URL

packages/data-store/src/__tests__/contact.store.test.ts
```diff
@@ -367,6 +367,15 @@ describe('Contact store tests', (): void => {
...
+          identifier: {
+            type: CorrelationIdentifierType.DID,
+            correlationId: 'example_did4',
```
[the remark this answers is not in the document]
BtencateSphereon: changed to URL

packages/oid4vci-holder/src/machine/oid4vciMachine.ts
sanderPostma: I don't see why
```
console.log('alice' && 'bob'); // Outputs: 'bob'
console.log('' && 'bob');      // Outputs: ''
console.log('alice' && '');    // Outputs: ''
```

BtencateSphereon: as discussed, it was only evaluating the last one. so for now i just combined the guards into one. one guard was not really doing anything yet anyways

packages/oid4vci-holder/src/agent/OID4VCIHolder.ts
BtencateSphereon: fixed, it now only returns true if there are results in the array

packages/oid4vci-holder/src/agent/OID4VCIHolder.ts
BtencateSphereon: not exposed anymore

packages/oid4vci-holder/src/agent/OID4VCIHolder.ts
BtencateSphereon: all this function is is a wrapper around calling the resolve function. so i guess i am not getting what you mean here. the resolve function needs a url, and trustAnchors. it is being called here

sanderPostma (approved):
LGTM

PR #261, opened 2024-10-31
chore: sdJwtDecodedCredentialToUniformCredential fix

sanderPostma:
Load credentialSubject from decodedPayload and not using decoded.disclosures

### DIFF
--- a/packages/ssi-types/src/types/sd-jwt-vc.ts
+++ b/packages/ssi-types/src/types/sd-jwt-vc.ts
@@ -324,15 +324,13 @@ export const sdJwtDecodedCredentialToUniformCredential = (
     return acc
   }, {} as Record<string, any>)
 
-  // Add id to credentialSubject if applicable
-  if (!credentialSubject.id && (sub || jti)) {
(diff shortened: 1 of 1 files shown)


packages/ssi-types/src/types/sd-jwt-vc.ts
```diff
@@ -325,13 +310,29 @@ export const sdJwtDecodedCredentialToUniformCredential = (
...
+      !(typeof value === 'object' && value !== null && Object.keys(value).length === 0)
+    ) {
+      acc[key] = value
+    }
+    return acc
+  }, {} as Record<string, any>)
+
+  // Add id to credentialSubject if applicable
+  if (sub || jti) {
+    credentialSubject.id = sub ?? jti
```
nklomp, line 329: What if id was already set by the reducer?

nklomp (changes requested):

packages/ssi-types/src/types/sd-jwt-vc.ts
sanderPostma: fixed

PR #262, opened 2024-10-31
fix: mdoc parsing

TimoGlastra:
Updates the mdoc parsing logic to be more compatible with how wrapped credentials behave for the other credential formats (especially for PEX)

The decoded payload on a wrapped mdoc was currently the class instance of the mdoc, but PEX uses the `decoded` payload to filter based on the paths in the PD.

So i updated the `decoded` credential to instead be a resolved JSON object where the keys are the namespaces and the values are the decoded JSON values. This is enough for pex to automticallly support the filtering of mdoc credentials. (tested it in PEX already).

Because the `decoded` now doesn't contain the mdoc class instance anymore I've added this  as the `credential` (this is similar how we do it with SD-JWT: decoded is just decoded payload for PEX matching, and credential is whatever that credential format wants to use to represent it).

I also removed the usage of the `mdocDecodedCredentialToUniformCredential(mdoc)` method, as it seemed very focused on the PID / not generally applicable. You can still get this by calling get uniform credential method.

As I mentioned in slack there are some issues with extracting the device signed, which for the PEX logic is not important, but it is important for the OID4vP integration to extract the nonce. But this can be addressed in a separate PR.

Also added some test vectors from other libs.

I'll make a follow up PR in PEX with the needed changes.


sanderPostma (approved):
LGTM

PR #266, opened 2024-11-07
fixes/SPHEREON-1185_FIDS

sanderPostma:


packages/oid4vci-holder/src/agent/OID4VCIHolder.ts
```diff
@@ -227,8 +227,8 @@ export class OID4VCIHolder implements IAgentPlugin {
     // "JcsEd25519Signature2020"
   ]
   private readonly didMethodPreferences: Array<SupportedDidMethodEnum> = [
+    SupportedDidMethodEnum.DID_JWK, // FIXME prefer JWK until we devise a method to detect when to use EBSI/jcs for did:key and when not
```
sanderPostma, line 230: Not sure whether this is ok for now

packages/oid4vci-holder/src/agent/OID4VCIHolderService.ts
```diff
@@ -261,14 +261,14 @@ export const getIdentifierOpts = async (args: GetIdentifierArgs): Promise<Manage
...
     })
-    if (created) {
-      await agentContext.agent.emit(OID4VCIHolderEvent.IDENTIFIER_CREATED, { result })
-    }
     identifier = await context.agent.identifierManagedGetByDid({
       identifier: result,
       keyType,
       offlineWhenNoDIDRegistered: result.did.startsWith('did:ebsi:'),
     })
+    if (created) {
```
BtencateSphereon, line 269: we were passing in a value called result. what happened to that?

packages/siopv2-oid4vp-op-auth/src/session/OpSession.ts
```diff
@@ -309,8 +309,12 @@ export class OpSession {
       const totalInputDescriptors = request.presentationDefinitions?.reduce((sum, pd) => {
         return sum + pd.definition.input_descriptors.length
       }, 0)
+      const totalVCs = args.verifiablePresentations?.reduce((sum, vp) => {
+        const uvp = CredentialMapper.toUniformPresentation(vp)
+        return sum + (uvp.verifiableCredential?.length ?? 0)
+      }, 0)
 
-      if (!request.presentationDefinitions || !args.verifiablePresentations || args.verifiablePresentations.length !== totalInputDescriptors) {
+      if (!request.presentationDefinitions || !args.verifiablePresentations || totalVCs !== totalInputDescriptors) {
```
BtencateSphereon, line 317: InputDescriptor does not equal 1 vc?

BtencateSphereon (changes requested):

packages/oid4vci-holder/src/agent/OID4VCIHolderService.ts
sanderPostma: The went inidentifierManagedGetByDid
The mobile wallet expects a managed identifier now, not a string

packages/siopv2-oid4vp-op-auth/src/session/OpSession.ts
sanderPostma: I guess so, but each VP can have one or more VCs. So two VCs can be in one VP and a third can be in another VP.
"args.verifiablePresentations.length !== totalInputDescriptors" is not handling that.

BtencateSphereon (approved):

PR #267, opened 2024-11-07
feature/SPRIND-83

zoemaas:


packages/anomaly-detection/src/types/IAnomalyDetection.ts
```diff
@@ -0,0 +1,16 @@
+import {IAgentContext, IPluginMethodMap} from '@veramo/core'
+import {AnomalyDetectionStoreArgs, IAnomalyDetectionStore} from "@sphereon/ssi-sdk.anomaly-detection-store";
+import {Optional} from "nx/src/project-graph/plugins";
```
sanderPostma, line 3: /src import

sanderPostma (changes requested):
Found bad import

sanderPostma (approved):
LGTM when import is fixed.
When this is not possible for some reason, add a comment

packages/anomaly-detection/src/types/IAnomalyDetection.ts
zoemaas: It looks like a mixed it up with Java :see_no_evil:

zoemaas: Replaced it with a Partial

packages/anomaly-detection-store/README.md
```diff
@@ -0,0 +1,55 @@
...
+  <br>
+</h1>
+
+---
+
+**Warning: This package still is in very early development. Breaking changes without notice will happen at this point!**
+
+---
+
+A Veramo anomaly detection store plugin. This plugin stores the result of the location lookup which is part of the anomaly detection plugin
```
nklomp, line 15: Please stop using the word Veramo everywhere. This really is not a Veramo plugin

packages/anomaly-detection-store/README.md
```diff
@@ -0,0 +1,55 @@
...
+## Usage
+
+### Adding the plugin to an agent:
+
+```typescript
+import { AnomalyDetectionStore } from '@sphereon/ssi-sdk.anomaly-detection'
+
+const agent = createAgent<IAnomalyDetectionStore>({
+  plugins: [
+    new AnomalyDetectionStore({
```
nklomp, line 30: This should be renamed to GeoLocationStore

packages/anomaly-detection-store/README.md
```diff
@@ -0,0 +1,55 @@
...
+      geoIpDBPath: './GeoLite2-Country.mmdb',
+    }),
+  ],
+})
+```
+
+### Lookup a location:
+
+```typescript
+const result = await agent.lookupLocation({
```
nklomp, line 40: Agent plugin should prefix something, to ensure that a developer can actually make head or tails about the 100+ methods available in a plugin.

So geoLookupLocation makes sense

packages/anomaly-detection-store/src/agent/AnomalyDetectionStore.ts
```diff
@@ -0,0 +1,140 @@
...
+  'anomalyDetectionStoreRemoveLocation',
+  'anomalyDetectionStoreClearAllLocations',
+  'anomalyDetectionStoreGetLocation',
+  'anomalyDetectionStoreDefaultLocationStore'
+]
+
+/**
+ * {@inheritDoc IAnomalyDetectionStore}
+ */
+export class AnomalyDetectionStore implements IAgentPlugin {
```
nklomp, line 28: Please rename to GeoLocationStore. This is only about Geo locations, which happens to be part of anomaly detection. We will not store other data in this plugin in the future. 

So rename the class and also this store plugin. But do not rename the anomaly-detection plugin itself. That just uses the renamed GeoLocationStore module/plugin

packages/anomaly-detection-store/README.md
zoemaas: Replaced Veramo with Sphereon

packages/anomaly-detection-store/README.md
zoemaas: geolocation is a single word: https://dictionary.cambridge.org/dictionary/english/geolocation. If we want to prefix the method with geo, doesn't it make more sense to call it geolocationLookup?

packages/anomaly-detection-store/README.md
zoemaas: geolocation is a single word: https://dictionary.cambridge.org/dictionary/english/geolocation. Doesn't it make more sense to rename it to GeolocationStore?

packages/anomaly-detection-store/README.md
nklomp: yes

packages/anomaly-detection-store/src/agent/AnomalyDetectionStore.ts
nklomp: of course

PR #269, opened 2024-11-07
Feature/sprind 93 oidf support

sanderPostma:
DRAFT
Needs to be rebased to develop, but I need wrap up & merge  FIDES-integration first


packages/kv-store/src/keyv/keyv.ts
```diff
@@ -156,6 +158,39 @@ export class Keyv<Value = any> extends EventEmitter implements KeyvStore<Value>
       .then(() => Promise.all(results))
   }
 
+  private isMapWithEntries(store: any): store is Map<string, Value> {
+    return store instanceof Map && typeof store.entries === 'function'
+  }
+
+  private async getFromMapWithPrefix(
```
sanderPostma, line 165: Not sure whether this is the best way to do it, but I need to fetch all key-values in a certain namespace (like you can de with the golang kv stores like badgerdb) without breaking all existing unit tests.

PR #276, opened 2024-11-13
Feature/sdk 44

jcmelati:


packages/oidf-client/__tests__/shared/oidfClientAgentLogic.ts
```diff
@@ -1,46 +1,31 @@
 import 'cross-fetch/polyfill'
```
sanderPostma, line 1: Can it still find this? "cross-fetch": "^3.1.8" is not in devDependencies anymore. 
(Maybe do "pnpm prune" followed by "pnpm build:clean" then run the test to confirm.

packages/oidf-client/src/agent/OIDFClient.ts
```diff
@@ -1,81 +1,70 @@
...
-import { schema } from '../index'
-import { JWK } from '@sphereon/ssi-types'
-import FederationClient = com.sphereon.oid.fed.client.FederationClient
-import DefaultFetchJSImpl = com.sphereon.oid.fed.client.fetch.DefaultFetchJSImpl
-import DefaultTrustChainJSImpl = com.sphereon.oid.fed.client.trustchain.DefaultTrustChainJSImpl
-import DefaultCallbacks = com.sphereon.oid.fed.client.service.DefaultCallbacks
+import {FederationClient, ICryptoService, IFetchService} from '@sphereon/openid-federation-client'
+import {JWK} from "@sphereon/ssi-types";
+import {IAgentPlugin} from '@veramo/core'
+import {Request} from "cross-fetch";
```
sanderPostma, line 4: Can it still find this? "cross-fetch": "^3.1.8" is not in devDependencies anymore. 
(Maybe do "pnpm prune" followed by "pnpm build:clean" then run the test to confirm.

packages/oidf-client/src/agent/OIDFClient.ts
```diff
@@ -1,81 +1,70 @@
...
+          input: requestInfo,
+          resourceType: 'application/entity-statement+jwt',
         })
-        DefaultCallbacks.setFetchServiceDefault(new DefaultFetchJSImpl())
-        DefaultCallbacks.setTrustChainServiceDefault(new DefaultTrustChainJSImpl())
-        this.oidfClient = new FederationClient()
-      } catch (error) {
-        throw Error(`Could not initialize the federation client: ${error.message}`)
+
+        return await response.text()
```
sanderPostma, line 51: What are we doing with the HTTP status >= 300?  response.text() may return an error message or a blank string.

sanderPostma (changes requested):

packages/oidf-client/src/agent/OIDFClient.ts
jcmelati: I added a check to throw if status is not 200.

packages/oidf-client/package.json
```diff
@@ -1,5 +1,6 @@
 {
 	"name": "@sphereon/ssi-sdk.oidf-client",
+	"type": "module",
```
sanderPostma, line 3: @jcmelati 
setting type to module specifies the type of modules that the project will use. and this is the only one with this option. and from my experience is most likely never the solution to the problem

sanderPostma (changes requested):

sanderPostma (approved):

PR #280, opened 2024-11-20
feature/SDK-41_oidf-support

sanderPostma:


packages/siopv2-oid4vp-rp-auth/src/types/ISIOPv2RP.ts
```diff
@@ -193,6 +195,7 @@ export interface IPresentationWithDefinition {
 export interface ISIOPIdentifierOptions extends Omit<IDIDOptions, 'idOpts'> {
   // we replace the legacy idOpts with the Managed Identifier opts from the identifier resolution module
   idOpts: ManagedIdentifierOptsOrResult
+  oidfOpts?: ExternalIdentifierOIDFEntityIdOpts
```
sanderPostma, line 198: Took a shortcut here by keeping it side-by-side

packages/oidf-client/package.json
```diff
@@ -1,5 +1,6 @@
 {
 	"name": "@sphereon/ssi-sdk.oidf-client",
+	"type": "module",
```
BtencateSphereon, line 3: hmm is this a change we should make?

packages/tsconfig.json
```diff
@@ -48,6 +48,6 @@
     { "path": "ebsi-support" },
     { "path": "public-key-hosting" },
     { "path": "resource-resolver" },
-	{ "path":  "oidf-client" }
+	{ "path": "oidf-client" }
```
BtencateSphereon, line 51: missing oidf-metatdata-server?

packages/kv-store/src/keyv/keyv.ts
```diff
@@ -156,6 +158,39 @@ export class Keyv<Value = any> extends EventEmitter implements KeyvStore<Value>
       .then(() => Promise.all(results))
   }
 
+  private isMapWithEntries(store: any): store is Map<string, Value> {
+    return store instanceof Map && typeof store.entries === 'function'
+  }
+
+  private async getFromMapWithPrefix(
+    prefixes: Array<string>,
+    options?: {
```
BtencateSphereon, line 167: should we use this?

packages/kv-store/src/keyv/keyv.ts
```diff
@@ -156,6 +158,39 @@ export class Keyv<Value = any> extends EventEmitter implements KeyvStore<Value>
...
+  ): Promise<Array<KeyvStoredData<Value>>> {
+    if (!this.isMapWithEntries(this.store)) {
+      return []
+    }
+
+    const map = this.store as Map<string, Value>
+    const entries = Array.from(map.entries())
+
+    const result = []
+    for (const prefix of prefixes) {
```
BtencateSphereon, line 179: can we not clean this stuff up with a reduce?

something like this

const map = this.store as Map<string, Value>;
const entries = Array.from(map.entries());

const result = prefixes.reduce((acc, prefix) => {
  const matchingEntry = entries.find(([key]) => key.startsWith(prefix));
  if (matchingEntry) {
    acc.push(matchingEntry[1] as KeyvStoredData<Value>);
  } else {
    acc.push(undefined);
  }
  return acc;
}, []);

return result;

packages/oid4vci-issuer-store/src/agent/OID4VCIStore.ts
```diff
@@ -84,17 +90,19 @@ export class OID4VCIStore implements IAgentPlugin {
         }),
       )
     }
+
     if (opts && Array.isArray(opts?.importMetadatas)) {
-      opts.importMetadatas.forEach((meta) =>
-        this.oid4vciStorePersistMetadata({
+      opts.importMetadatas.forEach((metaImport: IMetadataImportArgs) => {
+        const meta = metaImport as IIssuerMetadataImportArgs
```
BtencateSphereon, line 96: is this cast safe?

BtencateSphereon (changes requested):

packages/oidf-client/package.json
sanderPostma: No idea, it's from John's branch SDK-44 which I needed for the new oidf-client. He must have added it for a reason

packages/tsconfig.json
sanderPostma: fixed

packages/kv-store/src/keyv/keyv.ts
sanderPostma: removed, this started as a copy from get()

packages/oidf-client/package.json
BtencateSphereon: so this PR is now merging in SDK-44? this is not what we should do then? this branch should then have been based on SDK-44?

BtencateSphereon: also setting type to module specifies the type of modules that the project will use. and this is the only one with this option. and from my experience is most likely never the solution to the problem

sanderPostma: Fixed, SDK-44 is merged also, then I synched up with develop

BtencateSphereon (approved):

PR #282, opened 2024-11-22
feature/SPRIND-83

zoemaas:


nklomp (changes requested):
Why doesn't this change impact the anomaly detection plugin?

zoemaas:
@nklomp Could you please elaborate?  Many files were impacted by this change in the anomaly-detection plugin.

nklomp:
You have the geo location store and you have the anomaly detection plugin that is using the store. The geolocation plugin has been renamed. Why doesn't that impact any code in the anomaly detection plugin?

PR #297, opened 2024-12-19
feature/VDX-341

sanderPostma:
**BEWARE** this is based against SPRIND-137 so do not merge to develop until that one is merged,


packages/data-store/src/__tests__/pd-manager.entities.test.ts
```diff
@@ -24,6 +24,83 @@ describe('PresentationDefinitionItemEntity tests', (): void => {
...
+  it('should update PresentationDefinitionItemEntity dcqlPayload', async (): Promise<void> => {
+    const repository = dbConnection.getRepository(PresentationDefinitionItemEntity)
+    const entity = new PresentationDefinitionItemEntity()
+    entity.definitionId = 'definition1'
+    entity.version = '1.0'
+    entity.definitionPayload = JSON.stringify({ id: 'definition1', input_descriptors: [] })
+    entity.dcqlPayload = JSON.stringify({
+      credentials: [
+        {
+          id: 'credential1',
```
ronal2do, line 69: is there any kinda of validation on this ID? for example can it be duplicated or null?

packages/vc-handler-ld-local/src/suites/impl/JsonWebSignatureWithRSASupport.ts
```diff
@@ -11,6 +11,10 @@ import * as u8a from 'uint8arrays'
 import { RequiredAgentMethods } from '../../ld-suites'
 
 import { JsonWebKey } from './JsonWebKeyWithRSASupport'
+import { VerificationMethod } from '@sphereon/ssi-types'
+import { JsonWebKey2020, P256Key2021, P384Key2021, P521Key2021 } from '@transmute/web-crypto-key-pair'
+import { Ed25519VerificationKey2018 } from '@transmute/ed25519-key-pair'
+import { EcdsaSecp256k1VerificationKey2019 } from '@transmute/secp256k1-key-pair'
 
 // import { getJwaAlgFromJwk } from '@transmute/web-crypto-key-pair/dist/signatures/jws'
```
ronal2do, line 19: can it be removed?

packages/data-store/src/types/presentationDefinition/presentationDefinition.ts
```diff
@@ -8,10 +9,11 @@ export type PresentationDefinitionItem = {
...
   definitionPayload: IPresentationDefinition
+  dcqlPayload?: DcqlQuery
   createdAt: Date
   lastUpdatedAt: Date
 }
 
 export type NonPersistedPresentationDefinitionItem = Omit<PresentationDefinitionItem, 'id' | 'createdAt' | 'lastUpdatedAt'>
 export type PartialPresentationDefinitionItem = Partial<PresentationDefinitionItem>
-export type PresentationDefinitionItemFilter = Partial<Omit<PresentationDefinitionItem, 'definitionPayload'>> // TODO add logic to linearize & hash definitionPayload into a separate column so we can filter on it?
+export type PresentationDefinitionItemFilter = Partial<Omit<PresentationDefinitionItem, 'definitionPayload' | 'dcqlPayload'>> // TODO add logic to linearize & hash definitionPayload into a separate column so we can filter on it?
```
ronal2do, line 19: can we move this todo to a ticket and link it on the comment? it can easily be lost on the code

packages/data-store/src/__tests__/pd-manager.entities.test.ts
sanderPostma: The object in question is a query, the field can be anything or be omitted just like could be done qith SQL queries

packages/vc-handler-ld-local/src/suites/impl/JsonWebSignatureWithRSASupport.ts
sanderPostma: It's not my line / not commented out by me.
image

I typically do not remove other people's comments. Perhaps this one is very trivial, maybe it's something that still has to be hooked up in the future, I wouldn't know..
We can create a rule to clean up commented out imports, but until then I do not touch them.

packages/data-store/src/types/presentationDefinition/presentationDefinition.ts
sanderPostma: Actually I changed my mind and removed the TODO. The same presentation definition payload can be used for multiple configurations and retrieving records by payload hash may not return the intended record or multiple records, so better to avoid it. The caller just needs to the either the unique ID or definitionId + tenantId + version to query the correct record.

PR #298, opened 2025-01-09
chore: help application/x-www-form-urlencoded decoder to reconstruct vp_token of type object

sanderPostma:


zoemaas (changes requested):
Tests missing

sanderPostma:
> Tests missing

package siopv2-oid4vp-rp-rest-api does not provider jest tests. There is something else under __tests__ which is not explained how that is supposed to work.

The code is tested in a live agent.

zoemaas (approved):
LGTM

PR #299, opened 2025-01-09
feature/SDK-59

sanderPostma:


packages/vc-handler-ld-local/src/ld-credential-module.ts
```diff
@@ -115,7 +115,7 @@ export class LdCredentialModule {
     verificationMethodId: string,
     challenge: string | undefined,
     domain: string | undefined,
-    purpose: typeof ProofPurpose = !challenge && !domain
+    purpose: typeof ProofPurpose = !challenge // domain is not a mandatory var for AuthenticationProofPurpose per se, but challenge is.
```
ronal2do, line 118: I only miss to have unit tests for this module, to verify regressions

ronal2do (approved):

packages/vc-handler-ld-local/src/ld-credential-module.ts
nklomp: Has this change been verified to be correct to begin with in the specs @sanderPostma? The fact the underlying Veramo functions allow for it doesn't mean this change is correct. I cannot remember a need ever to not have both

nklomp (changes requested):
Needs to be checked against spec, but this change is dangerous in itself as it would start promoting not using domain bindings for proofs

PR #300, opened 2025-01-09
feature/SPRIND-89

BtencateSphereon:


packages/oid4vci-issuer/src/types/IOID4VCIIssuer.ts
```diff
@@ -82,6 +82,12 @@ export interface IIssuerOptions {
   didOpts?: IDIDOptions
   userPinRequired?: boolean
   cNonceExpiresIn?: number
+
+  /**
+   * Used in the callbacks for the first party flow
+   */
+  // FIXME we need to start supporting a map with a definition id per credential, we can use the credential offer session to check which credential is being issued and then look it up in this map
```
nklomp, line 89: Please link a ticket for this

packages/siopv2-oid4vp-rp-rest-api/src/siop-api-functions.ts
```diff
@@ -88,8 +89,15 @@ export function verifyAuthResponseSIOPv2Endpoint(
         // const credentialSubject = wrappedPresentation.presentation.verifiableCredential[0]?.credential?.credentialSubject
         // console.log(JSON.stringify(credentialSubject, null, 2))
         console.log('PRESENTATION:' + JSON.stringify(wrappedPresentation.presentation, null, 2))
-        const responseRedirectURI = await context.agent.siopGetRedirectURI({ correlationId, definitionId, state: verifiedResponse.state })
         response.statusCode = 200
+
+        const authorizationChallengeValidationResponse: AuthorizationChallengeValidationResponse = { presentation_during_issuance_session: verifiedResponse.correlationId }
+        if (authorizationResponse.is_first_party) {
```
nklomp, line 95: This looks spec incompliant and thus incompatible with any other wallet. Should be derived from metadata and/or certain attributes in responses instead of having our own prop in the response

nklomp (changes requested):
Small remarks. Similar like in the low-level lib about the is first party being tucked in the response.

packages/oid4vci-issuer/src/types/IOID4VCIIssuer.ts
BtencateSphereon: added the ticket number to this FIXME

packages/siopv2-oid4vp-rp-rest-api/src/siop-api-functions.ts
BtencateSphereon: created a ticket for this https://sphereon.atlassian.net/browse/MWALL-721

nklomp (approved):
LGTM

PR #301, opened 2025-01-10
chore: prepare interface for OAuth2StatusList args

sanderPostma:


packages/vc-status-list/src/types/index.ts
```diff
@@ -23,33 +23,58 @@ import {
...
+}
+
+export interface UpdateStatusListFromEncodedListArgs {
+  type?: StatusListType
+  statusListIndex: number | string
+  value: boolean
   proofFormat?: ProofFormat
   keyRef?: string
+  correlationId?: string
+  statusList2021Args?: UpdateStatusList2021Args
```
nklomp, line 69: Don;t call the name Args

packages/vc-status-list/src/types/index.ts
```diff
@@ -23,33 +23,58 @@ import {
...
+  indexingDirection: StatusListIndexingDirection
+  statusPurpose?: StatusPurpose2021
+  // todo: validFrom and validUntil
 }
 
-export interface UpdateStatusListFromEncodedListArgs extends StatusList2021ToVerifiableCredentialArgs {
-  statusListIndex: number | string
-  value: boolean
+export type OAuth2StatusListArgs = {
+  status: 'active' | 'suspended' | 'revoked'
```
nklomp, line 34: Make this more in line with the StatusList2021Args. So statusPurpose and extract the type

packages/vc-status-list/src/types/index.ts
```diff
@@ -23,33 +23,58 @@ import {
...
-  value: boolean
+  statusList2021Args?: StatusList2021Args
+  oauth2StatusListArgs?: OAuth2StatusListArgs
 }
 
-export interface StatusList2021ToVerifiableCredentialArgs {
+export type UpdateStatusList2021Args = {
   issuer: string | IIssuer
   id: string
-  type?: StatusListType
```
nklomp, line 46: Have you double checked why it was there? In other words can it really be removed?

packages/vc-status-list/src/types/index.ts
```diff
@@ -23,33 +23,58 @@ import {
...
   statusPurpose: StatusPurpose2021
   encodedList: string
+}
+
+export type UpdateOAuth2StatusListArgs = {
+  status: 'active' | 'suspended' | 'revoked'
```
nklomp, line 58: Please keep it symetrical with StatusList2021. So statusPurpose and extract the values. The fact you now already have the same values multiple times should have been an indicator

packages/vc-status-list/src/types/index.ts
```diff
@@ -23,33 +23,58 @@ import {
...
+  proofFormat?: ProofFormat
   keyRef?: string
-  statusListIndex: number | string
-  value: boolean
+  statusList2021Args?: StatusList2021Args
+  oauth2StatusListArgs?: OAuth2StatusListArgs
 }
 
-export interface StatusList2021ToVerifiableCredentialArgs {
+export type UpdateStatusList2021Args = {
```
nklomp, line 50: This is a pretty odd namechange. Are you sure that is correct.

The interface seems to be used from the name for a function that actually creates the VerifiableCredential from the status list

packages/ssi-types/src/types/w3c-vc.ts
```diff
@@ -278,6 +278,7 @@ export interface IErrorDetails {
 
 export enum StatusListType {
   StatusList2021 = 'StatusList2021',
+  OAuth2StatusList = 'OAuth2StatusList',
```
nklomp, line 281: This should not end up in w3c-vc for now. OAuth is not compatible

packages/vc-status-list/src/types/index.ts
sanderPostma: StatusList2021ToVerifiableCredentialArgs got lost, but put it back now

packages/ssi-types/src/types/w3c-vc.ts
sanderPostma: fixed

packages/vc-status-list/src/types/index.ts
```diff
@@ -23,33 +23,58 @@ import {
...
+  type: StatusListType
+  id: string
+  issuer: string | IIssuer
+  correlationId?: string
+  length?: number
+  proofFormat?: ProofFormat
   keyRef?: string
-  statusListIndex: number | string
-  value: boolean
+  statusList2021Args?: StatusList2021Args
```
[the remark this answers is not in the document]
sanderPostma: fixed

packages/vc-status-list/src/types/index.ts
sanderPostma: It's a different type, StatusList2021ToVerifiableCredentialArgs got lost but is back now.

packages/vc-status-list/src/types/index.ts
sanderPostma: fixed

packages/vc-status-list/src/types/index.ts
sanderPostma: fixed

packages/vc-status-list/src/types/index.ts
sanderPostma: fixed

PR #309, opened 2025-02-05
feature/IATAB2B-52

sanderPostma:


packages/vc-status-list-issuer-rest-api/src/api-functions.ts
```diff
@@ -96,6 +102,84 @@ export function getStatusListCredentialIndexStatusEndpoint(router: Router, conte
...
+      if (!entry) {
+        entry = {
+          statusList: details.id,
+          value: '0',
+          statusListIndex: resultStatusIndex,
+        }
+      }
+
+      response.statusCode = 200
+      return response.send({ ...entry, status })
```
nklomp, line 171: response.json. To make sure we always have the correct headers

packages/vc-status-list-issuer-rest-api/src/api-functions.ts
```diff
@@ -96,6 +102,84 @@ export function getStatusListCredentialIndexStatusEndpoint(router: Router, conte
     console.log(`Get statusList credential index status endpoint is disabled`)
     return
   }
+
+  const path = opts?.path ?? '/status-lists/:statusListId/entry/:entryId'
```
nklomp, line 106: Keep it more similar to the other one '/status-lists/:statusListId/status/entry-by-id/:entryId

The choice to have /status in there was on purpose

packages/vc-status-list-issuer-rest-api/src/api-functions.ts
```diff
@@ -187,13 +275,9 @@ export function updateStatusEndpoint(router: Router, context: IRequiredContext,
...
       } else {
-        if (!updateRequest.statusListIndex || updateRequest.statusListIndex < 0) {
-          return sendErrorResponse(response, 400, 'Invalid statusListIndex supplied')
-        }
         statusListEntry = await driver.getStatusListEntryByIndex({
-          statusListIndex: updateRequest.statusListIndex,
-          ...(statusListId && { statusListId }),
-          ...(!statusListId && { correlationId: statusListCorrelationId }), // Do not query for correlationId when we have statusListId
+          statusListId: details.id,
+          ...(entryCorrelationId ? { correlationId: entryCorrelationId } : { statusListIndex: updateRequest.statusListIndex }),
```
nklomp, line 280: Why is this passing an entryCorrelationId to a statusList correlation id?

packages/vc-status-list-issuer/src/agent/StatusListPlugin.ts
```diff
@@ -117,56 +143,129 @@ export class StatusListPlugin implements IAgentPlugin {
...
+    logger.debug(`Adding status to credential ${credential.id ?? 'without ID'}`)
+
     const credentialStatus = credential.credentialStatus
-    if (!credentialStatus) {
-      logger.info(`Not adding status list info, since no credentialStatus object was present in the credential`)
-      return Promise.resolve(credential)
+    if (credentialStatus) {
+      let existingStatusId: string | undefined
+      if (Array.isArray(credentialStatus)) {
+        // This was implemented with VC v2.0 support, but the rest of the SDK is not ready for that, so ICredential.credentialStatus's array union is disabled for now
```
nklomp, line 164: vc 2.0? We do not support that. You mean VCDM

nklomp (changes requested):
See remarks

packages/data-store/src/types/statusList/IAbstractStatusListStore.ts
```diff
@@ -11,9 +11,9 @@ export interface IStatusListEntryAvailableArgs {
 }
 
 export interface IGetStatusListEntryByIndexArgs {
-  statusListId?: string
+  statusListId: string
```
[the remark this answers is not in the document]
sanderPostma: In StatusListEntryEntity correlationId is the entity correlationId, not the status list correlationId.
I can make statusListId optional, no problem, but then the correlation id must be unique over all status lists.

packages/vc-status-list-issuer-rest-api/src/api-functions.ts
sanderPostma: fixed (though response.send does detect it's JSON atm.)

packages/vc-status-list-issuer-rest-api/src/api-functions.ts
sanderPostma: adjusted

packages/vc-status-list-issuer-rest-api/src/api-functions.ts
sanderPostma: [PERMALINK](https://github.com/Sphereon-Opensource/ssi-sdk/blob/ba7b1c04da999658e9172c86331bc0d0da784562/packages/data-store/src/types/statusList/statusList.ts#L35-L47)
There is only one correlationId so I assumed the correlationId is for the Entry and not for the list.

Shall I create two fields for this in the DB entity? statusListCorrelationId and entryCorrelationId  ?

packages/vc-status-list-issuer/src/agent/StatusListPlugin.ts
sanderPostma: Yes I mean this one
https://www.w3.org/TR/vc-data-model-2.0/#example-use-of-multiple-entries-for-the-status-property

packages/data-store/src/types/statusList/IAbstractStatusListStore.ts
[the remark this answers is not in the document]
nklomp: Not entirely sure what you mean but there are always 2 correlationIds of course that you can use instead of the id of the list and/or the index of the entry. They are the correlationId for the list instead of the id, and the correlationId of the entry instead of the index.

packages/vc-status-list-issuer-rest-api/src/api-functions.ts
nklomp: No, we have seen examples where this returned html all of a sudden

packages/vc-status-list-issuer-rest-api/src/api-functions.ts
nklomp: Yes, see other remark. The above certainly is wrong

packages/vc-status-list-issuer/src/agent/StatusListPlugin.ts
nklomp: So vcdm indeed

packages/data-store/src/types/statusList/IAbstractStatusListStore.ts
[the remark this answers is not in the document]
sanderPostma: So I have made statusListId required because it's the primary key for table StatusListEntryEntity and sole link between StatusListEntity and StatusListEntryEntity.
The StatusListEntryEntity has a correlation id field, but I assumed that as corralation id for the entry, not the list and it's not a unique column so it can have duplicate values for different status lists I suppose.

packages/vc-status-list-issuer-rest-api/src/api-functions.ts
sanderPostma: Ok, so we do need two correlation id fields in StatusListEntryEntity. Will have to rename both columns then.

sanderPostma: fixed

packages/data-store/src/types/statusList/IAbstractStatusListStore.ts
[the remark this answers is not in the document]
sanderPostma: Reverted, you can now filter on both status list correlationId or status list id. Since statusList has a @ManyToOne I can pass in both filters

packages/vc-status-list-issuer-rest-api/src/api-functions.ts
```diff
@@ -160,8 +244,8 @@ export function updateStatusEndpoint(router: Router, context: IRequiredContext,
     try {
       debug(JSON.stringify(request.body, null, 2))
       const updateRequest = request.body as UpdateW3cCredentialStatusRequest | UpdateIndexedCredentialStatusRequest
-      const statusListId = updateRequest.statusListId ?? request.query.statusListId?.toString() ?? opts.statusListId
-      const statusListCorrelationId = updateRequest.statusListCorrelationId ?? request.query.statusListorrelationId?.toString() ?? opts.correlationId
+      const statusListId = updateRequest.statusListId ?? request.query.statusListId?.toString() ?? opts.statusListId // TODO why query params when we have a JSON body ??
```
zoemaas, line 247: Do we have a ticket for that?

zoemaas (changes requested):
Left some comments

nklomp:
Already merged

PR #324, opened 2025-08-15
Feature/merge crypto extensions modules

nklomp:


sanderPostma (approved):
LGTM
Since nearly all code came from crypto-ext, most code lines are imported and unchanged.
The few things I found were already in teh old repo.

The library was tested in the mobile wallet with @sphereon/ssi-sdk-ext.*": "0.34.1-feature.SSISDK.13.51 artifacts and were confirmed working for VCI.

I did encounter some ESM related FIXMEs, please make sure they are registered in Jira or elsewhere if we still need to address them.

PR #330, opened 2025-09-08
feature/SSISDK-47

BtencateSphereon:


sanderPostma (changes requested):
Question, are clients like siopv2-oid4vp-rp-rest-client updated to use the new /backend/auth endpoints? Or do we need to use different APIs now?
UPDATE
-> Must be SSISDK-48

sanderPostma (approved):

PR #394, opened 2026-03-19
chore/VDX-59_credential_design_manager_fixes

BtencateSphereon:


packages/credential-design-manager/src/agent/CredentialDesignManager.ts
```diff
@@ -41,28 +58,201 @@ export class CredentialDesignManager implements IAgentPlugin {
...
 
   /** {@inheritDoc ICredentialDesignManager.cdmRemoveCredentialDesign} */
-  private async cdmRemoveCredentialDesign(args: RemoveCredentialDesignArgs, context: RequiredContext): Promise<boolean> {
-    await this.store.removeCredentialDesign(args)
-    return true
+  private async cdmRemoveCredentialDesign(args: RemoveCredentialDesignArgs, context: RequiredContext): Promise<RemoveCredentialDesignResult> {
+    return this.store
+      .removeCredentialDesign({ credentialDesignId: args.credentialDesignId })
+      .then(() => ({ result: true }))
+      .catch(() => ({ result: false }))
```
sanderPostma: Shouldn't we rethrow unexpected errors (!= NotFound) 
Like connection failures, constraint violations, permission errors

packages/data-store/src/credentialDesign/CredentialDesignStore.ts
```diff
@@ -122,14 +147,29 @@ export class CredentialDesignStore extends AbstractCredentialDesignStore {
...
 
-      const saved = await transactionalEntityManager.save(MetaDataSetEntity, existing)
+      const saved = await transactionalEntityManager.save(MetadataSetEntity, existing)
       return credentialDesignFrom(saved)
     })
   }
 
+  formStepGetOrCreate = async (args: FormStepGetOrCreateArgs): Promise<string> => {
+    debug('formStepGetOrCreate', args)
+    const repo: Repository<FormStepEntity> = (await this.dbConnection).getRepository(FormStepEntity)
```
sanderPostma, line 157: Why not dataSource.transaction() like the other updates?

sanderPostma (changes requested):
Few questions

packages/credential-design-manager/src/agent/CredentialDesignManager.ts
sanderPostma: If we do not do this anywhere also, it's a thing to pay attention to when port the thing to Kotlin

packages/data-store/src/credentialDesign/CredentialDesignStore.ts
BtencateSphereon: because there is nothing to put in a transaction. it is a single action

packages/credential-design-manager/src/agent/CredentialDesignManager.ts
BtencateSphereon: removed the catch for now to let veramo rest api handle the response.

sanderPostma (approved):
LGTM

PR #398, opened 2026-04-23
fix: request entity too large

sanderPostma:


packages/ssi-express-support/src/express-builders.ts
```diff
@@ -261,8 +270,9 @@ export class ExpressBuilder {
     // @ts-ignore
     opts?.handlers && app.use(opts.handlers)
     //fixme: this should come from the config
-    app.use(bodyParser.urlencoded({ extended: true }))
-    app.use(bodyParser.json({ limit: '5mb' }))
+    const limit = this._maxRequestBodySize ?? '5mb'
```
BtencateSphereon: should this default not be configurable?

BtencateSphereon (changes requested):
rfc

packages/ssi-express-support/src/express-builders.ts
sanderPostma: It is configurable through the builder withMaxRequestBodySize. That is the whole point for this change.
5mb It's the bodyParser library default which I exceeded

